PR #6, opened 2023-09-19
Chore: Fix e2e tests.

TimOrtel:
After the past few changes, the e2e tests started failing.


docker/e2e-tests.yml
```diff
@@ -15,14 +15,15 @@ services:
...
-      - DEFAULT_TIMEOUT=10000
+      - DEFAULT_TIMEOUT=30000
       - ADMIN_USERNAME=artemis_admin
       - ADMIN_PASSWORD=artemis_admin
       - robolectric.logging.enabled=true
     command:
       - test
       - -Dskip.debugVariants=true
       - -Dskip.flavor.unrestricted=true
+      - -Dslip.flavor.beta=true
```
Strohgelaender, line 26:
```suggestion
      - -Dskip.flavor.beta=true
```

TimOrtel: Good catch. I fixed it!

PR #22, opened 2023-11-22
Feature: Chat reference link support

TimOrtel:
[chat_links.webm](https://github.com/ls1intum/artemis-android/assets/100865202/f0fa30e5-2fe7-49b1-a5c2-d5e74ab2da9c)

User can now click on user mentions to directly open the chat with them. If the chat does not exist yet, it will be created in the background. Also supports opening conversation links.


feature/metis/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/ui/NavigateToUserConversationUi.kt
```diff
@@ -0,0 +1,175 @@
...
+                authToken,
+                serverUrl
+            )
+        }
+    }
+        .stateIn(viewModelScope, SharingStarted.Eagerly)
+
+
+    /**
+     * Load the conversation id associated with this username. If none exists, we create a new conversatin
```
nityanandaz, line 83:
```suggestion
     * Load the conversation id associated with this username. If none exists, we create a new conversation
```

PR #49, opened 2024-10-07
`Chore`: Update compose navigation and use type safe navigation

TimOrtel:
- Updated compose libraries
- Replaced route based navigation with object based navigation


asliayk (approved):
lgtm

FelberMartin (approved):
Also lgtm, can be merged

FelberMartin:
After merging the commit, I found that it causes the app to crash at startup for me. Is this only a problem at my side? @TimOrtel 
I reverted the merge to keep the main branch stable.

Here is the error that I am getting:
image

PR #58, opened 2024-10-18
Fix thread answers not shown

FelberMartin:
Fixed bug causing answers in threads to not display.

This PR solves the issue [#53](https://github.com/ls1intum/artemis-android/issues/53)

PR #70, opened 2024-10-24
`Communication`: Reintroduce foreign key constraints and offline primary keys

### LINKED ISSUE #62: `Communication`: Index constraints cause crashes
Recently a [hotfix](https://github.com/ls1intum/artemis-android/releases/tag/1.0.1) was released that removed some of the foreign key constraints. This was due to the foreign key constraints causing a variety of crashes in the application. However, this means that the data integrity is not maintained properly.

The foreign key and primary key constraints should be reworked in such a way that they:

1. Do not cause any crashes
2. Maintain proper data integrity

PaRangger:
### Motivation

In the hotfix for version [1.0.1](https://github.com/ls1intum/artemis-android/releases/tag/1.0.1) and [1.0.2](https://github.com/ls1intum/artemis-android/releases/tag/1.0.2) the foreign key constraints for reactions were removed. Additionally, the primary key constraints for posts and reactions were altered to include the server_id. These changes were done to prevent the application from crashing. This however introduced the problem that while the app was running without crashes, the data integrity is not given anymore.

### Description

Pull Request #58 fixed an issue with the answer threads not being shown. This issue also seemed to be the origin of the failing primary and foreign key constraints. Therefore I reverted the changes to include the previous FK and PK checks as they were.

Closes #62

### DIFF
--- a/feature/metis/shared/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/shared/db/entities/MetisPostContextEntity.kt
+++ b/feature/metis/shared/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/shared/db/entities/MetisPostContextEntity.kt
@@ -14,7 +14,7 @@ import androidx.room.Index
  */
 @Entity(
     tableName = "metis_post_context",
-    primaryKeys = ["client_post_id", "course_id", "conversation_id", "type", "server_id"],
+    primaryKeys = ["client_post_id", "course_id", "conversation_id", "type"],
     foreignKeys = [
         ForeignKey(
             entity = BasePostingEntity::class,
--- a/feature/metis/shared/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/shared/db/entities/MetisPostContextEntity.kt
+++ b/feature/metis/shared/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/shared/db/entities/MetisPostContextEntity.kt
@@ -14,7 +14,7 @@ import androidx.room.Index
  */
 @Entity(
     tableName = "metis_post_context",
-    primaryKeys = ["client_post_id", "course_id", "conversation_id", "type"],
+    primaryKeys = ["client_post_id", "course_id", "conversation_id", "type", "server_id"],
     foreignKeys = [
         ForeignKey(
             entity = BasePostingEntity::class,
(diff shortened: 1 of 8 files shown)


FelberMartin (approved):
Tested the App and works as expected without any crashes.

Also tested the migration from the 1.0.3 release to this app state, everything works fine 👍

feature/metis/shared/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/shared/db/entities/MetisPostContextEntity.kt
```diff
@@ -14,7 +14,7 @@ import androidx.room.Index
  */
 @Entity(
     tableName = "metis_post_context",
-    primaryKeys = ["client_post_id", "course_id", "conversation_id", "type", "server_id"],
```
TimOrtel, line 17: Why is the server id here not needed?

feature/metis/shared/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/shared/db/entities/PostReactionEntity.kt
```diff
@@ -7,7 +7,20 @@ import androidx.room.Index
 
 @Entity(
     tableName = "reactions",
-    primaryKeys = ["post_id", "emoji", "author_id", "server_id"],
```
TimOrtel, line 10: Why is the server id here not needed?

feature/metis/shared/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/shared/db/entities/MetisPostContextEntity.kt
PaRangger: Hey, it used to be like this pre hotfix. As far as I know from reading through comments in the code the server_id could be null in an offline-state.

feature/metis/shared/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/shared/db/entities/PostReactionEntity.kt
PaRangger: Same as above 👍🏼

julian-wls (approved):
Just tested it, and everything seems to work fine.

TimOrtel (changes requested):
Please add some unit Tests to verify that the referential integrity works by writing tests that add messages, delete messages, delete conversations etc.

julian-wls (approved):
Ran MetisDaoTest and tested adding posts, deleting posts and post reactions. Worked as expected, reapprove.

Issue #72, opened 2024-10-24
`Communication`: Autocompletion Improvement for Tagging Users and Channels


FelberMartin:
### Bug

**Current behaviour**
When typing a tagging ("#" or "@") character and my search term, the autocompletion list is shown as expected. After typing a space character, the list is still shown and I have to close it manually by tapping outside of the list.
![image](https://github.com/user-attachments/assets/0faa4600-f944-45a0-b0e9-d4788f882b9a)

**Expected behaviour** 
After typing a space, the autocompletion list should not be shown anymore.

### UI Enhancement

The autocompletion list currently is partly drawn over the Preview and Edit buttons and could visually be more separated from the rest of the UI.
![image](https://github.com/user-attachments/assets/dfba1a2d-7db4-4210-a4f6-c9b2d90c4b14)

FelberMartin:
I just looked into the code and found the behaviour that I described as a bug, seems to be intentional. At least the code for this communicates this. https://github.com/ls1intum/artemis-android/blob/7a2f61d5ea1d05ab58850bab8322d5abc97a826b/feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/reply/ReplyTextField.kt#L625-L660
 @TimOrtel Do you know what the reason for this behaviour was? 

In the WebApp the Autocompletion disappears after typing the first space. On iOS the behaviour is different again, there the Autocompletion does not disappear at all until you click a entry from the list. But I assume this can be considered a bug (@anian03) ?

anian03:
> In the WebApp the Autocompletion disappears after typing the first space. On iOS the behaviour is different again, there the Autocompletion does not disappear at all until you click a entry from the list. But I assume this can be considered a bug (@anian03) ?

From what I know on iOS this was intentional at the time because we had no way of knowing where the cursor is, and we assumed the user still wanted to pick an exercise. But I agree with you that we should also consider changing this on iOS as well.

TimOrtel:
Pretty sure it is because when tagging people they have a space between first name and last name.

FelberMartin:
After going to sleep I just realised that, thanks for the hint :D

PR #76, opened 2024-10-25
`Communication`: Improve auto completion for tagging users and channels

FelberMartin:
Fixed the clipping of the AutoCompletionPopup as described in #72 and closes this issue.

Additionally also changed the search for courses, lectures, and channels to be case-insensitive to be more intuitive.

Before:
image

After:
image


TimOrtel:
Can you please add screenshots on how this will look in the following cases:
1. entering a multi line text into the text field. e.g. when already being in line 3 and then starting an autocomplete dialog
2. when doing the same on a foldable
3. when doing the same on a tablet

FelberMartin:
> Can you please add screenshots on how this will look in the following cases:
> 
>     1. entering a multi line text into the text field. e.g. when already being in line 3 and then starting an autocomplete dialog
>     2. when doing the same on a foldable
>     3. when doing the same on a tablet

I am not at my dev setup right now, so I can't provide screenshots at the moment.

Regarding (1.), with my current understanding of the code, the popup will show above the edit / preview buttons, regardless of the cursor position. I will check that as soon as I am back.

Regarding the foldable and tablet, dont know what the behaviour would be like and I understand that this is important for future proving the app. However, I feel like adjusting the UI for tablet and foldables should probably be part of a different bigger ticket (as this likely also includes a completely altered design (with sidebar navigation eg)) and we should focus our limited resources on the main form factor for now.

TimOrtel:
> > Can you please add screenshots on how this will look in the following cases:
> > 
> >     1. entering a multi line text into the text field. e.g. when already being in line 3 and then starting an autocomplete dialog
> >     2. when doing the same on a foldable
> >     3. when doing the same on a tablet
> 
> I am not at my dev setup right now, so I can't provide screenshots at the moment.
> 
> Regarding (1.), with my current understanding of the code, the popup will show above the edit / preview buttons, regardless of the cursor position. I will check that as soon as I am back.
> 
> Regarding the foldable and tablet, dont know what the behaviour would be like and I understand that this is important for future proving the app. However, I feel like adjusting the UI for tablet and foldables should probably be part of a different bigger ticket (as this likely also includes a completely altered design (with sidebar navigation eg)) and we should focus our limited resources on the main form factor for now.

The app, and especially the communication component already supports tablet mode interactions. It was part of a greater feature I implemented a some time ago. Therefore, it is important that when making these changes we consider how the tablet and foldable layouts are affected. 

So please provide the screenshots once you have time.

FelberMartin:
Sorry, I did not know that tablet mode is already supported. Here are the screenshots:

Multiline mode on Phone:
image

Foldable:
image


Autocompletion on Tablet:
image

@TimOrtel Do you think the autocompletion dialog is fine for the tablet, or should we introduce some kind of maximum width, so the dialog does not span the whole screen?

julian-wls (approved):
Just tested it and it looks good to me. The changes seem reasonable.

I think introducing some kind of maximum width does not feel right as it would require users to stretch their fingers all across the screen to reach the middle, which might be annoying.
Although it feels strange to have the view covering the entire screen, I think it might lead to a better user experience.

FelberMartin:
Ah sorry, I meant adding a maximum height for tablets, not width. My fault

julian-wls:
I think a maximum height would be quite reasonable. It would definitely look visually more appealing on tablets. We could set it to be the height of 5-6 items.

FelberMartin:
### New Changes 
- I added a maxHeight to the dialog to show about 5-6 items max
- I also updated the design to be more consistent

### Screenshots
#### Phone
image
image

#### Foldable
image

#### Tablet
image

julian-wls (changes requested):
I just tested it. This looks already so much better. The idea of introducing a max height was really good. Code lgtm.

There is only one thing I noticed:
The dividers are not visible in dark theme:
Screenshot4

Does this happen to you as well?
I think we could fix this as part of this PR.

FelberMartin:
Good catch, I'll have a look

FelberMartin:
Okay, so it turns out this is a flaw of the [M3 design specifications](https://m3.material.io/styles/color/static/baseline#d3170e61-484c-4c35-a847-2aae11803ccb). Specifically, in dark mode, the colors `surfaceVariant` and `outlineVariant` are exactly the same ("neutral-variant30"). This causes all dividers to be invisible in dark mode.

We could either:

1. Ignore the problem
2. Deviate from the M3 design specs and manually change the value for `md_theme_dark_outlineVariant` in `Color.kt`
3. Not use `surfaceVariant` at all and replace it with something else in the app to avoid the problem. In a newer versions of M3 than we currently use, there eg is `surfaceContainer`.  (This would be a separate PR for sure)

I would suggest that for now we either do 1 or 2 (I'd tent to 2), and eventually follow up with 3 separately, as this involves updating the theme and replacing all occurrences of `surfaceVariant`. What do you think? @julian-wls

julian-wls:
> I would suggest that for now we either do 1 or 2 (I'd tent to 2), and eventually follow up with 3 separately, as this involves updating the theme and replacing all occurrences of `surfaceVariant`. What do you think? @julian-wls

I agree. Let's do 2 and follow up with 3 in the future as part of a different PR :)

FelberMartin:
Turns out the app uses dynamic colors for sdk >= 31, so we can not directly change the outlineVariant color. So, these changes will only be visible for users with sdk 30. 

I also noticed that lifting the requirement for minSdk from 30 -> 31 was not necassary (as no other dependencies require this), so I lowered it again to 30.

I created a follow up issue #102

julian-wls (approved):

PR #77, opened 2024-10-29
`Chore`: Fix End To End Tests Containers

TimOrtel:
- Fixed end to end test setup locally to get them running again
- Updated README on how to start the end to end tests locally


FelberMartin (approved):
Tested it locally, and works, thank you!

There are a few tests failing (or is it just for me?), but I assume we should fix this in a separate ticket.

PR #88, opened 2024-11-03
Improve TextField IME actions

FelberMartin:
### Problem
Currently many TextFields show a return / newLine key to the user, even for TextFields where this does not make sense (eg for entering names, URLs, emails, etc.)

### Change description
I scanned for all TextFields in the app and added the appropriate IME actions to them.


julian-wls (approved):
Just tested it. The changes seem appropriate and I didn't notice any issues when using the textfields.

PR #91, opened 2024-11-07
Display content edge-to-edge

### LINKED ISSUE #68: Re-Enabling EdgeToEdge
I found that with `enableEdgeToEdge()` introduced in #43 the keyboard was covering the message TextField in the Chat UI (see screenshot below).
image

The same also holds for other TextFields. Eg when tapping on the Search in the ConversationOverview, the Keyboard covers the FAB-Button (for creating new chats). Without the `enableEdgeToEdge()` this does not happen.

Therefore I temporally removed this line from `main`. 

### Next Steps
- Confirm whether this is not just an issue on the emulator but also on the physical device.
- Resolve the issue. I found that by adding `Modifer.imePadding()` eg to line 320 of `CourseUiScreen.kt` does remove the problem. However, I am not sure what the best approach is to fix this consistently. There are probably much more TextFields that cause the same problem and adding the Modifier to each and every Scaffold does not feel right imo

julian-wls:
### Description
This PR introduces edge-to-edge content, which will be enforced with Android 15 (SDK 35).
The changes were made by following the official [documentation](https://developer.android.com/develop/ui/compose/layouts/insets).
For lazy lists, the content now goes all the way to the edge of the screen leading to the new intended way to display content. Therefore the bottom padding has been replaced with content padding. imePadding has been introduced for items, that normally stick to the end of the screen, but should still be visible, when the ime changes its size.
This PR closes #68.

### Testing
The changes have been tested on foldables, tablets and both screen orientations.

### Screenshots
Content consumes system bar:
Screenshot2
Content padding to keep content in the safe area:
Screenshot3


julian-wls:
> * When clicking on a lecture, the Lecture Units items are cut off at the bottom of the screen
> * CreateChat: Clicking the FAB and "Create Chat", the info (eg "Please enter at least 3 characters ...") is hidden behind the keyboard

Thanks, for pointing that out, I'll go ahead and make the required changes.

> Also, have you tried to adding an UI test that asserts that the ReplyTextField is still visibile after the Keyboard popped up? I am not sure whether compose tests simulate the keyboard, though. I think we should definetly somehow ensure that the TextField is visible with the Keyboard popping up, as this is like the main usecase for our app.

That sounds like a really good idea. I'll try to investigate this further and see if I can implement another test class, that adds textfield visibility tests.

feature/metis/conversation/src/test/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/reply/ReplyTextFieldVisibilityUITest.kt
```diff
@@ -0,0 +1,168 @@
+package de.tum.informatics.www1.artemis.native_app.feature.metis.conversation.ui.reply
+
+import androidx.compose.foundation.layout.PaddingValues
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.imePadding
```
FelberMartin, line 5: There are many unused imports here, please remove them.
To prevent this in the future, I suggest to do this automatically when commiting by going to Settings > Version Control > Commit > Commit Checks > Optimize Imports

feature/metis/conversation/src/test/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/reply/ReplyTextFieldVisibilityUITest.kt
```diff
@@ -0,0 +1,168 @@
...
+
+        val newPosition = markdownTextField.fetchSemanticsNode().positionInRoot.y
+
+        markdownTextField
+            .assertExists()
+            .assertIsDisplayed()
+        assertTrue("Text field should move up when the keyboard appears", newPosition < initialPosition)
+    }
+
+    private fun setupThreadUi(post: PostPojo) {
```
FelberMartin, line 116: A very similar method exists in `ConversationAnswerMessagesUITest` maybe create common superclass (could also contain the "setup" of the post and answer pojos)

FelberMartin (changes requested):

feature/metis/conversation/src/test/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/reply/ReplyTextFieldVisibilityUITest.kt
julian-wls: Thanks for the hint! I just changed it :)

feature/metis/conversation/src/test/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/reply/ReplyTextFieldVisibilityUITest.kt
```diff
@@ -0,0 +1,84 @@
...
+import de.tum.informatics.www1.artemis.native_app.feature.metis.conversation.ui.chatlist.PostsDataState
+import de.tum.informatics.www1.artemis.native_app.feature.metis.shared.db.pojo.PostPojo
+import junit.framework.TestCase.assertTrue
+import kotlinx.coroutines.CompletableDeferred
+import org.junit.Test
+import org.junit.experimental.categories.Category
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+
+@Category(EndToEndTest::class)
```
FelberMartin, line 26: UiTests can be run without the E2E container setup, so change it to `UnitTest::class`

feature/metis/conversation/src/test/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ConversationAnswerMessagesUITest.kt
```diff
@@ -1,105 +1,37 @@
...
 import kotlinx.coroutines.CompletableDeferred
-import kotlinx.coroutines.Deferred
-import kotlinx.coroutines.runBlocking
-import kotlinx.datetime.Clock
 import org.junit.Test
 import org.junit.experimental.categories.Category
 import org.junit.runner.RunWith
 import org.robolectric.RobolectricTestRunner
 
 @Category(EndToEndTest::class)
```
FelberMartin, line 22: UiTests can be run without the E2E container setup, so change it to `UnitTest::class`

feature/metis/conversation/src/test/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/reply/ReplyTextFieldVisibilityUITest.kt
```diff
@@ -0,0 +1,84 @@
...
+    fun `test GIVEN the thread view is shown containing one post and three answer posts WHEN the markdown text field is clicked THEN the keyboard is shown below the markdown text field`() {
+        setupThreadUi(posts[0]) { CompletableDeferred() }
+        runTest()
+    }
+
+    @Test
+    fun `test GIVEN the chat list containing three posts is shown WHEN the markdown text field is clicked THEN the keyboard is shown below the markdown text field`() {
+        setupChatUi(posts)
+        runTest()
+    }
```
FelberMartin, line 40: Tbh I am not sure whether this is possible with our current UI test setup, but can you look into whether we could also add another test where `enableEdgeToEdge()` is set for the tests? 
Because I checked whether your tests would have been able to catch the problem initially introduced by `enableEdgeToEdge` where the keyboard covered the textField. I reverted the padding and inset changes in `ConversationChatListScreen`, but the tests still pass

FelberMartin (changes requested):

feature/metis/conversation/src/test/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/reply/ReplyTextFieldVisibilityUITest.kt
julian-wls: That's indeed problematic. I did some research on this, but apparently, there is no way to simulate `enableEdgeToEdge()` in compose UI tests.

julian-wls: I created this issue #121 to make sure we don't forget about it and maybe come up with an alternative test setting.

FelberMartin (approved):
Lgtm now

PR #96, opened 2024-11-08
Fix answer post got duplicated

### LINKED ISSUE #92: `Communication`: Group channel message in thread showing twice
On the Android app, when opening a thread a sending a message, the message is shown twice in the thread. On the WebApp however, the message is only shown once, so this only seems to be a local issue with the Android app, and does not affect the other apps. 

When testing the same in a 1:1 chat, this behaviour did not occur.

### Screenshot
image

FelberMartin:
From time to time, the Websocket UPDATE event will occur before the `upgradeClientSideAnswerPost` method was called. This lead to a duplication of the answer post.

This is now fixed by removing the older post if necessary.

This closes #92.


TimOrtel (changes requested):

TimOrtel:
Please add unit tests on the level of MetosStorageServiceImpl that verify that this behavior now works correctly.

feature/metis/shared/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/shared/db/MetisDao.kt
```diff
@@ -140,7 +140,7 @@ interface MetisDao {
         postingType: BasePostingEntity.PostingType = BasePostingEntity.PostingType.STANDALONE
     )
 
-    @Query("delete from metis_post_context where client_post_id = :clientPostId")
+    @Query("delete from postings where id = :clientPostId")
```
julian-wls, line 143: Why is the query different now?

feature/metis/conversation/src/test/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/service/storage/impl/MetisStorageServiceImplTest.kt
```diff
@@ -0,0 +1,171 @@
...
+
+        assertEquals(1, posts.size)
+        assertEquals(1, posts.first().answers.size)
+    }
+
+
+    @Test
+    fun testUpgradeClientSideAnswerPost() = runTest {
+        // Okay, this is super weird. Somehow the test sql database has problems if the test method
+        // name is too long. That's why I had to create this delegation method.
```
julian-wls, line 110: I'm not sure but in this particular case I would maybe even accept a less descriptive function title, even if it is not that straight forward. But in my opinion it's better than a weird delegation. What do you guys think @TimOrtel and @FelberMartin?

julian-wls (commented):
Everything seems reasonable, the tests make sense, the only thing I don't really understand is this change mentioned below. And I also add one remark regarding the delegation.
All in all, good work.

feature/metis/shared/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/shared/db/MetisDao.kt
FelberMartin: Good question, tbh I am not sure myself why this worked before. When using this method to delete answers it (understandably) caused the app to crash because of a foreign key error, as there now is a posting with the respective metis_post_context being deleted. I do not know why this worked previously for deleting standalone posts

feature/metis/conversation/src/test/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/service/storage/impl/MetisStorageServiceImplTest.kt
FelberMartin: Fair point, I could add the more descriptive parts to the respective comments ( `// GIVEN, WHEN, THEN`)

feature/metis/conversation/src/test/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/service/storage/impl/MetisStorageServiceImplTest.kt
```diff
@@ -0,0 +1,165 @@
...
+import kotlinx.datetime.Clock
+import org.junit.Assert.assertEquals
+import org.junit.Test
+import org.junit.experimental.categories.Category
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+
+@Category(UnitTest::class)
+@RunWith(RobolectricTestRunner::class)
+class MetisStorageServiceImplTest {
```
TimOrtel, line 25: In the future, we will have a lot more tests for this service. I suggest you rename this class to represent that the upgrade logic is tested in this class.

feature/metis/conversation/src/test/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/service/storage/impl/MetisStorageServiceImplTest.kt
```diff
@@ -0,0 +1,165 @@
...
+        tags = emptyList(),
+        answers = emptyList(),
+        reactions = emptyList()
+    )
+
+    private val basePost = StandalonePost(basePostPojo, conversation)
+    private val localAnswer = AnswerPost(localAnswerPojo, basePost)
+
+    @Test
+    fun testInsertClientSidePost() = runTest {
```
TimOrtel, line 80: Please add the test GIVEN WHEN THEN naming as a comment to the test functions if it does not work with the method names.

feature/metis/conversation/src/test/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/service/storage/impl/MetisStorageServiceImplTest.kt
```diff
@@ -0,0 +1,165 @@
...
+        )
+
+        // THEN: Both the base post and the answer post are stored
+        val posts = sut.getStoredPosts(
+            serverId = host,
+            metisContext = metisContext
+        ).loadAsList()
+
+        assertEquals(1, posts.size)
+        assertEquals(1, posts.first().answers.size)
```
TimOrtel, line 103: Please add assertions for the data within the retrieved posts. This just checks that a post with an answer is present, but not that they also hold identical data to what was previously inserted.

feature/metis/conversation/src/test/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/service/storage/impl/MetisStorageServiceImplTest.kt
```diff
@@ -0,0 +1,165 @@
...
+        )
+
+        // WHEN: insertOrUpdatePosts is called before upgradeClientSideAnswerPost.
+        val answerPojoUpdated = localAnswerPojo.copy(serverPostIdCache = localAnswerPojo.serverPostIdCache.copy(serverPostId = 1))
+        var basePostUpdated = StandalonePost(basePostPojo, conversation)
+        val answerUpdated = AnswerPost(answerPojoUpdated, basePostUpdated)
+        basePostUpdated = basePostUpdated.copy(answers = listOf(answerUpdated))
+
+        // Called by the WebSocket
+        sut.updatePost(
```
TimOrtel, line 129: There should be a test method for both orders.

feature/metis/conversation/src/test/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/service/storage/impl/MetisStorageServiceImplTest.kt
```diff
@@ -0,0 +1,165 @@
...
+            clientSidePostId = answerClientPostId,
+            post = answerUpdated
+        )
+
+        // THEN: Only one answer post is stored
+        val posts = sut.getStoredPosts(
+            serverId = host,
+            metisContext = metisContext
+        ).loadAsList()
+        assertEquals(1, posts.size)
```
TimOrtel, line 148: Please add checks on the content and that the post has actually been upgraded.

TimOrtel (changes requested):

feature/metis/conversation/src/test/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/service/storage/impl/MetisStorageServiceImplTest.kt
FelberMartin: As discussed in [this thread](https://github.com/ls1intum/artemis-android/pull/96#discussion_r1835414778), this leads to error in the SQL database, thats why I opted to go for the shorter method naming and instead add the more descriptive ones in the comments of the test

julian-wls (approved):
Code lgtm now!

feature/metis/conversation/src/test/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/service/storage/impl/MetisStorageServiceImplTestUpgradeLocalAnswerPost.kt
```diff
@@ -0,0 +1,211 @@
...
+        val posts = getStoredPosts()
+        assertEquals(answerUpdated.serverPostId, posts.first().answers.first().serverPostId)
+    }
+
+    private suspend fun getStoredPosts() = sut.getStoredPosts(
+        serverId = host,
+        metisContext = metisContext
+    ).loadAsList()
+
+    private suspend fun <T : Any>PagingSource<Int, T>.loadAsList(): List<T> {
```
TimOrtel, line 199: I am not sure if this function is implemented correctly, could you try adding the paging-testing library and replacing the implementation of this function with this? 

```kotlin
private suspend fun <T : Any> PagingSource<Int, T>.loadAsList(): List<T> {
    return Pager(PagingConfig(pageSize = 10), pagingSourceFactory = { this }).flow.asSnapshot {
        scrollTo(50)
    }
}
```

FelberMartin: Thank you for pointing that out, my implementation felt quite hacky indeed.

TimOrtel (approved):

TimOrtel:
Before merging, please check the tests. It seems like they have not been executed.

PR #59, opened 2024-10-19
`Communication`: Resolve Post functionality for threads

julian-wls:
### Description:
This PR includes the following features and fixes:
- A new PostAction in the PostActionContextBottomSheet that allows users to mark AnswerPosts in threads as resolving posts (Fig. 1)
- A new label on post items indicating if it is a resolved post or if it resolves a parent post (Fig. 2)
- Three minor fixes: 
	- Loads text into the markdown text field when editing post button has been clicked
	- Removes line limit of post items to allow longer messages to resolve #52 
	- Requests reload after successful completion of post modifications in the conversation view model to apply changes to the view

### Screenshots: 
Fig. 1:
Screenshot_20241019_103154

Fig. 2:
Screenshot_20241019_103212

If the post already resolves the parent posts users can remove this tag:
Screenshot_20241019_103223

PR #110, opened 2024-11-16
`Hotfix`: Fix post creation of MetisThreadUI to create replies instead of posts

julian-wls:
### Description
This PR is a **hotfix** that fixes an issue, which caused replies in the thread view not to be created correctly. Instead of calling the createReply function the createPost function of ConversationViewModel was called, leading to a post creation instead of a reply creation.
This bug only affects `develop` and has been introduced by the change of MetisThreadUI in #59.


FelberMartin (approved):
Works now

PR #114, opened 2024-11-18
`Communication`: Add ability to pin messenges

julian-wls:
### Description

This PR introduces pinning posts. A new post action has been created to allow users to pin standalone posts. Pinning posts is only available for channel creators and moderators, which is why the according flags are passed down to the post action. 

With the pinning feature added, it becomes quite clear that the post items in the chat need an overhaul to make the whole view coherent again. This will happen as a follow-up PR as described [here](https://github.com/ls1intum/artemis-android/issues/162).

### Changes

- The database had to be migrated as there was no column for the display priority in the Posts table.
- The PostItem was changed in order to give pinned posts a different look and show a pinned label on top of the post (see screenshot). 
- This PR includes multiple UI and E2E tests to ensure the correct behavior.
- MetisModificationService was changed to handle the modification request

### Steps for testing

1. Make sure you have moderation rights or create a chat before testing
2. Send a message and long click on it.
3. The `Pin Post` button should be visible.
4. Click the button and see how the message changes its appearance in the chat (A reload might be required because of ongoing issues with the websocket).
5. Long click again, click `Unpin Post` and verify that the message returns to its initial state
6. Verify that the Pin Post option is not available on answer posts in the thread view


feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/post/PostItem.kt
```diff
@@ -99,6 +106,7 @@ internal fun PostItem(
         PostItemViewType.ThreadContextPostItem -> true
         else -> false
     }
+    val isPinned = post is IStandalonePost && post.displayPriority == DisplayPriority.PINNED
```
FelberMartin, line 109: Is the condition `post is IStandalonePost` needed here? I feel like we should display the post as pinned according to the data status and the restriction that only standalone posts _can_ be pinned should be handled by the Server and the bottomSheet UI

feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/post/PostItem.kt
```diff
@@ -119,15 +127,36 @@ internal fun PostItem(
...
-                        onClick = onClick,
-                        onLongClick = onLongClick
-                    )
+                } else {
+                    it
+                        .let { modifier ->
+                            if (isPinned) {
+                                modifier
+                                    .clip(
+                                        RoundedCornerShape(8.dp)
```
FelberMartin, line 136: Please use MaterialTheme.shapes.medium (or small) instead

feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/post/PostItem.kt
```diff
@@ -119,15 +127,36 @@ internal fun PostItem(
                     it
```
FelberMartin, line 127: I think we should move the modifier modification outside to the function body for better readability, at the moment its quite a lot. Maybe just creating a function something like `val applyPinStatusToModifier: (Modifier) -> Modifier = { if (isPinned) ...` would be sufficient

feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/thread/MetisThreadUi.kt
```diff
@@ -287,7 +305,9 @@ private fun PostAndRepliesList(
                     verticalArrangement = Arrangement.spacedBy(8.dp)
                 ) {
                     PostWithBottomSheet(
-                        modifier = Modifier.padding(top = 8.dp),
+                        modifier = Modifier
+                            .padding(top = 8.dp)
+                            .testTag(testTagForPost(post.standalonePostId)),
```
FelberMartin, line 310: Notice that this will add the test tag also to the BottomSheet, I am not sure whether this is intended

feature/metis/conversation/src/test/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ConversationMessagesUITest.kt
```diff
@@ -0,0 +1,163 @@
...
+        )
+            .performSemanticsAction(SemanticsActions.OnLongClick)
+        composeTestRule.onNodeWithText(context.getString(R.string.post_pin))
+            .performClick()
+
+        testPinDestination(changedPost)
+    }
+
+    @Test
+    fun `test GIVEN post is not pinned WHEN pinning the post THEN the correct post gets unpinned`() {
```
FelberMartin, line 50: I assume this should be "... post is pinned WHEN unpinning ..."

feature/metis/conversation/src/test/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ConversationMessagesUITest.kt
```diff
@@ -0,0 +1,163 @@
...
+        )
+            .performSemanticsAction(SemanticsActions.OnLongClick)
+        composeTestRule.onNodeWithText(context.getString(R.string.post_pin))
+            .performClick()
+
+        testPinDestination(changedPost)
+    }
+
+    @Test
+    fun `test GIVEN post is not pinned in thread WHEN pinning the post THEN the correct post gets unpinned in thread`() {
```
FelberMartin, line 92: Same as above

feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/chatlist/MetisChatList.kt
```diff
@@ -96,6 +97,7 @@ internal fun MetisChatList(
         clientId = clientId,
         hasModerationRights = hasModerationRights,
         isAtLeastTutorInCourse = isAtLeastTutorInCourse,
+        isConversationCreator = isConversationCreator,
```
FelberMartin, line 100: At this point I feel like we maybe should wrap these flags in a wrapper class to reduce the clutter (as they are passed down a lot). But I cant really think of a suitable name, maybe something in the direction of `PostActionFlags`?

FelberMartin (changes requested):
Tested the functionality, works great! 

Only thing that I noticed, is that the web app allows users to pin messages in DMs, but this is not possible currently in Android

See code comments below

feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/thread/MetisThreadUi.kt
julian-wls: I think it's only added to the PostItem, since the modifier is not passed to PostContextBottomSheet (see PostWithBottomSheet.kt)

feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/post/PostItem.kt
FelberMartin: Ah, that makes sense. Then it's fine for me how it is 👍

FelberMartin (approved):
Pinning in DMs now also work, nice.

Just update the database version and its good to be merged from my side 👍

app/src/main/java/de/tum/informatics/www1/artemis/native_app/android/db/AppDatabase.kt
```diff
@@ -29,7 +29,7 @@ import de.tum.informatics.www1.artemis.native_app.feature.push.communication_not
         CommunicationMessageEntity::class
     ],
     exportSchema = true,
-    version = 10,
+    version = 11,
```
FelberMartin, line 32: With the latest release the database version was increased to 11, to update it to 12 here :)

ahbitaqu (approved):
Works perfectly over here!

PR #119, opened 2024-11-19
`Bugfix`: Removed moderator rights to edit posts; added test

FelberMartin:
### Problem Description
Currently a user with moderation rights can edit anyone's posts. As discussed in the Communication Team Meeting, this behaviour is unwanted.

### Changes
 
- Removed moderator rights to edit other's posts
- Created unit test

### Steps for testing
- Login as an instructor (has moderation rights)
- Long press another users post
- See that "Edit" is not available anymore
- Press on a post of yourself 
- "Edit" is still possible


julian-wls (approved):
Code lgtm. Works as expected!
As an instructor, I can only edit my own posts, but not others posts.

feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/post/PostActions.kt
```diff
@@ -46,24 +45,24 @@ fun rememberPostActions(
...
+            return@remember PostActions()
         }
+
+        val doesPostExistOnServer = post.serverPostId != null
+        val hasEditPostRights = post.authorId == clientId
+        val hasResolvePostRights = isAtLeastTutorInCourse || post.authorId == clientId
+
+        PostActions(
+            requestEditPost = if (doesPostExistOnServer && hasEditPostRights) onRequestEdit else null,
+            requestDeletePost = if (hasEditPostRights) onRequestDelete else null,
```
julian-wls, line 58: `hasEditPostRights` is used for deleting, too. However, deleting should be possible for moderators. Could you adjust this?

julian-wls (changes requested):
Sorry I should have looked better, I missed one crucial thing, see below.

feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/post/PostActions.kt
FelberMartin: Very true, good catch!

julian-wls (approved):

TimOrtel (changes requested):
The code looks good but I think it would be better if you added UI tests. We should have tests that actually view some posts, long click a post and assert that the edit button is available or not in the bottom sheet.

julian-wls:
I think the UI test would definitely make sense; I'm not sure, though, if we should include them as part of this PR. The changes were discussed in the last meeting and seem to be urgent. I would include them in the upcoming release and create an issue for the UI tests. Furthermore, there seems to be an ongoing issue with the RobolectricRunner and the modal bottom sheet, which is another reason to postpone the UI tests.

PR #124, opened 2024-11-20
`Bugfix`: Fix post insertion for live created posts 

### LINKED ISSUE #115: Message received in wrong channel
### Problem description
When receiving a new message while being in another chat in the android app, the new message is shown as belonging to the currently displayed chat. Even after reloading the message persists on the android app

In the following gif the android user is in the channel #tech-support and the webuser sends a message in channel #random. The android user sees the message in the #tech-support channel
ArtemisChatReceiveMessageInWrongChannel

### Next steps
- Fix bug
- Write test to prevent this from happening in the future

julian-wls:
### Problem Description
When having a channel open and receiving a post from another channel the post gets saved into the currently opened channel. This PR fixes this issue by inserting the correct conversationId into the database. This PR fixes #115 

### Changes
MetisStorageService has been updated to save the correct conversationId for the post.

### Steps for testing
1. Open a channel on the app
2. Send a message in a public channel on the website
3. Check if the message has not been saved in the currently opened channel on the app but in the correct one.

### DIFF
--- a/feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/service/storage/impl/MetisStorageServiceImpl.kt
+++ b/feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/service/storage/impl/MetisStorageServiceImpl.kt
@@ -146,12 +146,13 @@ internal class MetisStorageServiceImpl(
             serverId: String,
             clientSidePostId: String,
             serverSidePostId: Long?,
+            conversationId: Long?,
             postingType: BasePostingEntity.PostingType
         ): MetisPostContextEntity =
             MetisPostContextEntity(
                 serverId = serverId,
                 courseId = courseId,
-                conversationId = conversationId,
+                conversationId = conversationId ?: this.conversationId,
                 serverPostId = serverSidePostId,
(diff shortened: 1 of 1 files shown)


FelberMartin (changes requested):
Tested and the bug does not occur anymore.
Could you please add a test to ensure the proper functionality? You may use a similar structure as in the test introduced in https://github.com/ls1intum/artemis-android/pull/96

feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/service/storage/impl/MetisStorageServiceImpl.kt
```diff
@@ -146,12 +146,13 @@ internal class MetisStorageServiceImpl(
             serverId: String,
             clientSidePostId: String,
             serverSidePostId: Long?,
+            conversationId: Long?,
```
TimOrtel, line 149: I think this could would be cleaner if the parameter would be ```conversationId: Long = this.conversationId```. Then you do not need the ?: in the body.

TimOrtel (changes requested):

FelberMartin (approved):

PR #146, opened 2024-11-23
`Bugfix`: Fix manual refresh for post updates

FelberMartin:
### Problem Description
Currently the user has to manually refresh the conversation to show post updates (such as reactions, mark as resolved) for non-coursewide channels. 
This problem was caused by only subscribing to the coursewide websocket.

This PR will fix #60 

### Changes
 
- Subscribe to non-coursewide websocket
- Renamed parameter "channel" to "topic" to match iOS and Web implementations
- Introduced `WebsocketTopic` file to centrally manage the websocket topics

### Steps for testing
Test for both a coursewide channel (eg tech-support) and a non-coursewide channel (eg groupchat):
- Go to the channel
- Write a message -> is shown in chat
- React to the message -> reaction is shown without the need to manually refresh
- Reply to message in thread with new answer
- Mark answer as "Resolves Post" -> Answer is shown as resolving

### Follow up
As mentioned in the ConversationOverviewViewModel.kt, I noticed that the Websocket updates for the conversations are not received properly. I will take care of this in a direct follow up PR to not overload this PR with changes.

PR #147, opened 2024-11-24
`Bugfix`: Fix websocket conversation overview

FelberMartin:
### Problem Description
Conversation updates where not displayed via the websocket in the app. For example updating a conversation name did not reflect in the app. Also the unread messages counter did not increase for new incoming messages.

This is a follow up to #146.

### Changes
 
- Fix serialization of conversation updates -> updates such as renaming do now reflect in the app
- Also subscribed to postUpdate websocket to increment unread counters

### Steps for testing
Open a course both in the web app and go to the conversation overview in the android app:
- Web: Create new group chat with android user -> Android: see that group chat instantly appears 
- Web: Rename group chat -> Android: see that renaming is reflected without a manual refresh needed
- Web: Type a new message in the chat -> Android: see unread counter increases for chat


julian-wls (approved):
Changes work as expected, great work! Code lgtm.
There is one thing I noticed, though:
image

Apparently, the location of the badge is not set correctly. I'll open an issue for this as this is unrelated to this PR.

FelberMartin:
Okay if its introduced with this PR, then I'll try to fix it with this PR aswell

PR #167, opened 2024-11-29
`Bugfix`: Fix adding members to conversation

### LINKED ISSUE #139: [BUG] Error: Add member to chat
### Describe the bug
When trying to add a new member to a chat, an error occurs.

**To Reproduce**
1. Login into TS0 with test_user_1
2. Go to General Testing Course > Communication > Group Chats > #error-add-member
3. Click settings icon
4. Click "Add members"
5. Search for a user that is not yet in the group, eg "test user 4"
6. Click "Add"
7. Error dialog appears

**Expected behavior**
No error dialog should be shown and the user should be added to the chat.

### Screenshots
image

### Next steps
- Investigate where error root cause is
- Fix error
- Ensure proper behaviour with tests

julian-wls:
### Problem Description

There was an issue that prevented users from adding members to a conversation. Instead of adding the members, an error was shown.

### Changes

The cause of this error was a missing conversation id that ensures a match to the correct conversation. The manage_conversations_module has been changed to correctly set the conversationId parameter of the ConversationAddMembersViewModel. This PR fixes #139.
In Conversation AddMemberSettingsE2eTest, the functionality of adding members is tested. However, this bug did not affect the tests as the conversationId is set independently from the module there.

### Steps for testing

- Go to a group
- Click the settings icon
- Click `Add members`
- Select members, click `Add`
- Verify that the members get added correctly


FelberMartin (approved):
Thanks for fixing that issue!

It would be nice to also include a module initialization test to prevent this from happening again in the future, take `login_moduleTest` as a reference

feature/metis/manage-conversations/src/test/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/manageconversations/manage_conversations_moduleTest.kt
```diff
@@ -0,0 +1,121 @@
...
+            parametersOf(courseId, conversationId)
+        }
+    }
+
+    /**
+     * Test if a ViewModel can be initialized without errors.
+     * @param parameters The parameters to pass to the ViewModel constructor
+     */
+    @OptIn(KoinInternalApi::class)
+    private inline fun <reified T: ViewModel> ComposeContentTestRule.testViewModelInitialization(
```
FelberMartin, line 106: We should probably move this method to core/common-test if possible to avoid code duplication.

FelberMartin (changes requested):

feature/metis/manage-conversations/src/test/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/manageconversations/manage_conversations_moduleTest.kt
julian-wls: I tried moving it to BaseComposeTest, but there were errors, which is why I chose this solution.

PR #178, opened 2024-12-02
Redesign settings screen and remove clutter

### LINKED ISSUE #123: Redesign Settings screen
The settings screen should get an overhaul, so that 
- the different elements and sections are separated more clearly, 
- the layout is more consistent, 
- and clickable items are distinguishable from simple labels.

image

**Additional Info**
This issue was discovered in https://github.com/ls1intum/artemis-android/pull/122

julian-wls:
### Problem Description

The settings screen is currently quite cluttered, and it is difficult to distinguish between buttons and static text. The problems have been described in #123

### Changes

This PR changes the Preference sections and introduces new Preference entries to make the view more appealing and easier to understand. See the screenshots below. This PR fixes #123

### Screenshots

Before:

After:

### DIFF
--- a/feature/settings/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/settings/SettingsScreen.kt
+++ b/feature/settings/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/settings/SettingsScreen.kt
@@ -35,8 +35,8 @@ import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.rememberCoroutineScope
-import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.vector.ImageVector
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.text.font.FontWeight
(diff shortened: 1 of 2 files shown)


feature/settings/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/settings/SettingsScreen.kt
```diff
@@ -415,59 +418,149 @@ private fun BuildInformationSection(
...
-    Box(modifier = modifier.clickable(onClick = onClick)) {
-        Row(modifier = Modifier.padding(16.dp)) {
+fun ButtonEntry(
+    modifier: Modifier = Modifier,
+    text: String,
+    isFocused: Boolean = false,
+    textColor: Color = MaterialTheme.colorScheme.onSurface,
+    onClick: () -> Unit,
+) {
+    Box(
```
FelberMartin, line 517: Would it be sufficient here to remove the box and just have the row as the row layout? This also applies to the other Entries

FelberMartin (approved):
Looks really cool, just a small code comment

feature/settings/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/settings/SettingsScreen.kt
julian-wls: Yes, you're right. :+1:

PR #184, opened 2024-12-05
`Chore`: Remove unused bamboo buildNumber

FelberMartin:
### Problem Description
Stephan made me aware that we still have running bamboo builds which were replaced with github actions for this repo, so I also searched the code for remaining (and now unused) bamboo configs.

### Changes
 
Removed the bambo buildNumber.


julian-wls (approved):
Changes lgtm!

PR #189, opened 2024-12-06
`Bugfix`: Fix image jittering in chat for all images

### LINKED ISSUE #166: [BUG] Chatlist with small image jittering when reloading
### Describe the bug
#109 introduced displaying images. When reloading the chatlist with an image, it jitters. It seems to only happen for small images.

**To Reproduce**
**Big image -> no jittering** 
TS1 -> Practical Course: Interactive Learning WS24/25 -> #random

**Small image -> jittering**
TS1 -> Practical Course: Interactive Learning WS24/25 -> #tech-support

**Expected behavior**
No jittering also for small images

### Screenshots
ArtemisChatImageReloadJitter

### Additional context
See the discussion in #109

julian-wls:
### Problem Description

After applying #109 there was a problem that caused the chat to jitter when reloading. This was caused by images smaller than the default height, as described in #166.

### Changes

The root of the problem was unnecessary reloads that were triggered when scrolling through the list. Therefore, a condition has been introduced checking if the markdown text has changed and only performing an update if there are changes. This PR fixes #166

### Steps for testing

1. Go to a chat that contains images (preferably images which have a greater width than height)
2. Perform scrolling and reloading the chat
3. Verify that there is no jittering


FelberMartin (approved):
Awesome, now jittering anymore :D

In MarkdownText.kt in `createMarkdownRender`, there is still a comment about jittering and setting the image size. Is this still relevant?

ahbitaqu (commented):
I dont know if this is related to the issue and I can't really pin-point the exact circumstances, but I still get some jittering for only specific images, The majority seems to be working just fine...

https://github.com/user-attachments/assets/e85f2f57-cd08-4a1b-a9aa-9815a8b99ba5

julian-wls:
> I dont know if this is related to the issue and I can't really pin-point the exact circumstances, but I still get some jittering for only specific images, The majority seems to be working just fine...

Thanks for your review. I have seen this behavior before. I was able to reproduce it on `develop` but not on this branch. Since I applied the changes of this branch the problem was gone for me. Could you check again using only the changes introduced in this branch? Thanks for your help!

julian-wls:
> Awesome, now jittering anymore :D
> 
> In MarkdownText.kt in `createMarkdownRender`, there is still a comment about jittering and setting the image size. Is this still relevant? EDIT: @julian-wls Can you have a look at this? :)

I think setting the size should be kept. #192 utilizes the size setting to set the correct height for icons. I removed the comment though.

Issue #102, opened 2024-11-14
 Address Dark Mode Divider Visibility Issue by Updating Surface Colors


FelberMartin:
### Description

This follow-up issue to the problem discovered in [this PR discussion](https://github.com/ls1intum/artemis-android/pull/76#issuecomment-2466273070) aims to resolve the divider visibility problem in dark mode due to the M3 design specification flaw. Specifically, the colors `surfaceVariant `and `outlineVariant `are both mapped to `neutral-variant30`, making dividers invisible in dark mode.

In #76 we quickly fixed this issue by assigning a different value to `outlineVariant` for dark mode

### Tasks
- Dependency Update: Upgrade to the latest M3 version that includes surfaceContainer (or another suitable alternative) if available.
- Replace SurfaceVariant: Update all instances of surfaceVariant with surfaceContainer (or another suitable color).

PR #190, opened 2024-12-06
`Chore`: Disable dynmaic colors + replace surfaceVariant usages

FelberMartin:
### Problem Description
See #102 

### Changes
 
- Updated the theme related files with the latest version of the Figma Material 3 Theme builder.
- Disabled dynamic colors
- Replaced usages of `surfaceVariant` with `surfaceContainer`

### Steps for testing
- Start app
- See that dynamic colors are not applied anymore (eg when changing the wallpaper, the app colors do not change)
- In dark mode, go to any chat and start the AutoCompletionDialog by typing "#" or "@"
- See that the dividers between the suggestion items are visible

### Screenshots
image


julian-wls (approved):
Works as expected. Your changes are appropriate. 
Just one comment:

Is the blue border intended? I think it's not the primary blue, but a different one. We can leave it like this, I don't think it's a problem, I just wondered why this happened and wanted to point it out:
image

FelberMartin:
The search border indeed does not use the `primary`, but the `primaryContainer` color. But I think this is good, it should not draw too much attention imo

PR #197, opened 2024-12-06
`Chore`: Prepare websocket for compressed payload

### LINKED ISSUE #195: `Chore`: Prepare websockets for compressed payloads
### Problem
The Artemis server is currently facing performance problems due to too large websocket updates. Because of that, it is planned to automatically compress payloads that are larger than a certain threshold. This will we a breaking change and needs to be handled by the Android app websocket in advance.

### Next steps
- Check for `compressed: true` in header and de-compress the payload with Gzip
- Add tests

FelberMartin:
### Problem Description
To reduce the load on the server with sent websocket updates, the Artemis team decided to compress the payloads of websocket updates if the exceed a certain threshold.

This PR closes #195 

### Changes
 
- Added check for the websocket compression header and optionally decompress the websocket payload
- Added tests for both compressed and normal payloads

### Steps for testing
As the compression is not yet live on any servers, we can only test that un-compressed payloads work as before. 
- Go to any chat with the app
- Write a message to that chat from the webapp
- Observe that the new message is immediately displayed on the app (without the need for a refresh)


eylulnc (changes requested):
There is a problem with message deleting, I compared it with the develop and it is working fine there.

When I tried to delete a message I got the pop up for - Could not delete the message.

FelberMartin:
> I had a some related to message deleting but could not reproduced it after the second build. When I tried to delete a message I got the pop up for - Could not delete the message.

Okay that's weird, these websocket changes should actually have no effect on the deletion of messages🤔. In case you find a way to reproduce the bug, please also try whether it occurs on `develop` aswell. 

I also tried reproducing the error but did not get an error pop up.

julian-wls (commented):
I just tested the changes using TS1.
I experienced the following:
It worked for the first message and for the second message, the third message was not received, but the fourth was. For all other messages after that I always had to manually reload in order to fetch the message.

I checked the code, but I couldn't find a specific reason that could explain this behavior.

FelberMartin:
I do not think that the observed behaviours are related to this PR, I just tried to reproduce them again but it was able to observe the faulty behaviour. Additionally the compressed websocket payloads have not been enabled for the websocket topics affecting the native mobile apps yet. 

@julian-wls @eylulnc Would you mind trying to reproduce the faulty behaviours, and if they persist, also check whether they can also be observed on the `develop` branch? :)

julian-wls (approved):
I was not able to reproduce any of the issues I had before. My assumption is that the observed issues in my last review were related to network load. I also couldn't reproduce the issues on `develop`.
I don't think whatever the cause of this was is related to this PR.

Code lgtm.

PR #206, opened 2024-12-09
Redesign notification settings

### LINKED ISSUE #183: Notification settings redesign
### Problem

The notification settings are quite cluttered and overwhelming. #178 introduced a new design for the settings view, leading to an inconsistent look of these two views.

### Solution

The notification settings should be redesigned to match the new design of the settings view. Preference sections should be wrapped in a card design and the checkboxes should be replaced with switches.

julian-wls:
### Problem Description

Similar to the settings screen the notification screen is quite unstructured and cluttered.

### Changes

This PR introduces sections for preference entries in the form of cards and replaces checkboxes with switches to stay consistent with the design of the settings.
This PR closes #183.

### Screenshots

notificationSettings_new


feature/push/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/push/ui/PushNotificationSettingCategoriesListUi.kt
```diff
@@ -144,38 +162,40 @@ private fun PushNotificationSettingEntry(
...
-                    modifier = Modifier.testTag(TEST_TAG_PUSH_CHECK_BOX),
-                    isChecked = setting.push,
-                    text = stringResource(id = R.string.push_notification_settings_label_push),
-                    onCheckedChanged = { onUpdate(setting.webapp, setting.email, it) }
-                )
-            }
+        if (setting.push != null) {
+            Switch(
+                modifier = Modifier
+                    .scale(0.9f)
```
FelberMartin, line 171: Please introduce a constant for the scale value. This can then also be used in `PushNotificationSettingsUi`

FelberMartin (changes requested):
Looking much better, thank you for taking care of this!

Two concerns:
1. If the user changes a setting and navigates back, the settings will not get synced with the server and the user probably does not know this. Maybe we should add a dialog in case the user has unsaved changes.
2. I tried changing and saving some settings, but I very often get an error dialog (I think only the first save did not show an error dialog for me). This is the logcat output: 
image

FelberMartin:
I just noticed that it is currently possible to go to the Notification settings screen when the user is logged out, I assume this is not intended. Would you mind fixing this as part of this PR? @julian-wls 

I think it should be as easy as moving the `NotificationSection` into the `accountData?.let` statement in `SettingsScreen` :D

julian-wls:
> Looking much better, thank you for taking care of this!
> 
> Two concerns:
> 
> 1. If the user changes a setting and navigates back, the settings will not get synced with the server and the user probably does not know this. Maybe we should add a dialog in case the user has unsaved changes.
> 2. I tried changing and saving some settings, but I very often get an error dialog (I think only the first save did not show an error dialog for me). This is the logcat output:

1. I'll create a follow up PR tomorrow, which will probably include a dialog that warns a user when there are unsaved changes and has a button to save them.
2. I was not able to reproduce it. Correct me if I'm wrong:
       - I changed more than one switch
       - I clicked save
       - I did not receive an error in logcat

FelberMartin (approved):
Unfortunately I could also not reproduce the error consistently. But trying now I could only achieve it once, and by using the Network Inspector I saw that it was caused by a `500 Internal Server Error`, something about a duplicate key. However I missed to take a screenshot with the details and now I can not reproduce it anymore.

But anyways, as it is a server related error, this PR is good to merge :D

eylulnc (approved):
I also could not reproduce the already mentioned error log,  design looks good. 😊

PR #211, opened 2024-12-11
`Bugfix`: Close thread after deleting Parent Post

### LINKED ISSUE #209: [BUG] Deleting a post in the thread view causes loading error
### Describe the bug

After deleting a post in the thread view the user faces an error that indicates that the post couldn't be loaded as it has been removed from the database.

**To Reproduce**

1. Click on any post
2. Long-press the post
3. Click on delete Post
4. See the error

**Expected behavior**

The user should be redirected to chat instead of staying in the thread view facing an error.

### Screenshots

dd

### Additional context

This bug affects `develop`

julian-wls:
### Problem Description

As described in #209, after deleting the parent post the user stays in the thread view facing a loading error, because the post is not available anymore.

### Changes

The onNavigateUp method is now passed down to the view model to be invoked after the parent post got successfully deleted in the thread view. This PR closes #209.

### Steps for testing

1. Go to the thread view.
2. Delete the Parent Post
3. Verify that you get taken back to the previous screen.
4. Try the same without an internet connection and verify the error dialog is still shown.
5. Try deleting an answer post in the thread view and verify that you are not taken back to the previous screen
6. Try deleting a post outside of the thread view and verify it works like before.

### DIFF
--- a/feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/ConversationViewModel.kt
+++ b/feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/ConversationViewModel.kt
@@ -382,7 +382,7 @@ internal open class ConversationViewModel(
             )
                 .bind { if (it) null else MetisModificationFailure.DELETE_POST }
                 .or(MetisModificationFailure.DELETE_POST)
-                .also { if (it != MetisModificationFailure.DELETE_POST && post is IStandalonePost && onCloseThread != null) onCloseThread?.invoke() }
+                .also { if (it != MetisModificationFailure.DELETE_POST && post is IStandalonePost) onCloseThread?.invoke() }
         }
     }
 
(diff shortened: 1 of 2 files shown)


feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/ConversationViewModel.kt
```diff
@@ -380,6 +382,7 @@ internal open class ConversationViewModel(
             )
                 .bind { if (it) null else MetisModificationFailure.DELETE_POST }
                 .or(MetisModificationFailure.DELETE_POST)
+                .also { if (it != MetisModificationFailure.DELETE_POST && post is IStandalonePost && onCloseThread != null) onCloseThread?.invoke() }
```
FelberMartin, line 385: I think `&& onCloseThread != null` is not needed here due to the  null coalescing operator in the invoke call

FelberMartin (approved):
Works as described, good catch! See one minor code comment below

feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/ConversationViewModel.kt
julian-wls: Yes, you're right!

PR #222, opened 2024-12-13
`Feature`: Add navigation animations

FelberMartin:
### Problem Description
Currently there are no animations when navigating the app. There should be smooth animations when navigating from one screen to another.

### Changes
 
- Added navigation animations to all screens
- Adapted Server instance selection to show as BottomSheet instead of own screen

I have tested that adding the animations does not mess with the different layouts on both tablets and foldables.

### Steps for testing
Navigate through all the app's screens and see that there are now animations when navigating


feature/metis/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/ui/SinglePageConversationBody.kt
```diff
@@ -76,170 +85,144 @@ internal fun SinglePageConversationBody(
...
-        is AddChannelConfiguration -> {
-            if (canCreateChannel) {
-                CreateChannelScreen(
+            is BrowseChannelConfiguration -> {
+                BrowseChannelsScreen(
                     modifier = modifier,
                     courseId = courseId,
-                    onConversationCreated = openConversation,
+                    onNavigateToConversation = openConversation,
+                    //onNavigateToCreateChannel = {},
```
julian-wls, line 144: I assume you forgot to delete this :)

julian-wls (commented):
Just tested it. Navigating feels so much better now :)
I tested different animation durations and I think 220 is a good choice! I also like the new instance selection. Couldn't find any issues in the code, just one comment:

julian-wls (approved):

eylulnc (approved):
It’s looking good! 😊 The navigation feels smoother now, and the experience is much improved.

Issue #162, opened 2024-11-28
Redesign chat to be more comparable to iOS


julian-wls:
### Description

The chat looks rather outdated compared to iOS, which is mostly because the Post Items feel a bit cluttered and not visually appealing.
#114 already introduces post item cards for pinned posts. As a follow-up this should be the default. Furthermore, the time label should be placed in the corner and show the exact post time and date.

PR #228, opened 2024-12-16
`Communication`: Redesign chat & alignment with iOS

julian-wls:
### Problem Description

Currently, the chat is looking outdated and different from iOS, see #162.

### Changes

This PR applies the following changes to make the chat more readable and more aligned to the iOS design:
- Post Cards
- Role badges
- more information on edited posts
- show date on parent post in thread
- better padding management
- changed Date divider
- icon next to reply button
- added emoji button
- the post time has been moved to the top right corner


core/ui/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/core/ui/date/relativeTime.kt
```diff
@@ -27,7 +27,8 @@ fun getRelativeTime(
     to: Instant,
     clock: Clock = Clock.System,
     formatSeconds: Boolean = false,
-    showDate: Boolean = true
+    showDate: Boolean = true,
+    showDateAndTime: Boolean = false
 ): CharSequence {
```
FelberMartin, line 32: Im really struggling with this method and the boolean toggles tbh, I think its very hard to understand what the method emits in what situation. But I could also not find a truly satisfying solution for it. Like, what should I expect if I pass showDateAndTime = true, but showDate = false to the function? Maybe a enum? I really dont know. 
Overall the function just feels hard to read. Maybe a little bit of refactoring already helps, or maybe we just have to live with it, idk.
What are your thoughts on this?

FelberMartin (changes requested):
Tested the functionality on the device, looking much better now, thanks for taking care of this :D

Idk why, but the padding on the header feels weird to me, maybe its because the padding on the top is too small (and not consistent with the other paddings of the post)? But I also see that by increasing the padding there, we can fit even less posts onto the screen then. 
image

FelberMartin:
I just thought of this now, but could you also test that the changes work on Tablet and Foldable? :D

FelberMartin:
I just thought of this now, but could you also verify that these changes work on Tablet and Foldable aswell? :D

eylulnc (commented):
I agree with the padding adjustments Martin mentioned. While testing, I noticed an issue when replying to a message. Initially, the tag displays correctly, but after sending reply, it updates to an incorrect state and stays that way.

https://github.com/user-attachments/assets/5be518e3-692c-48a5-8f03-36bb417cbe9a

Screenshot 2024-12-22 at 15 28 48

We’ve observed a similar issue with images flickering due to re-rendering, but in this case, the state remains incorrect after the render. Should we investigate this as a separate issue, or could it be related to the existing follow-up ticket on re-render problems?

core/ui/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/core/ui/date/relativeTime.kt
julian-wls: I played around with it and tried different approaches, but there wasn't any satisfying option, enums were even harder to read imo. I think I would keep it as it is for now.

julian-wls:
> I agree with the padding adjustments Martin mentioned. While testing, I noticed an issue when replying to a message. Initially, the tag displays correctly, but after sending reply, it updates to an incorrect state and stays that way.

> We’ve observed a similar issue with images flickering due to re-rendering, but in this case, the state remains incorrect after the render. Should we investigate this as a separate issue, or could it be related to the existing follow-up ticket on re-render problems?


Yes, that's true. I also noticed this behavior, when implementing the changes. The bug is unrelated to this PR and affects `develop`, too. I already opened [this issue](https://github.com/ls1intum/artemis-android/issues/235) to tackle this bug.

julian-wls:
> I just thought of this now, but could you also verify that the changes work on Tablet and Foldable? :D

That's a good idea. I just tested the foldable and observed the following behavior:
CCCCC

As you can see the PostHeader is cut in the thread UI, which is caused by the small width. What's your opinion on this @FelberMartin? I'm not quite sure how we can solve this. Is there an option to align the chat and the thread in a way that the width of both views is the same? I think this would be better long-term solution as the thread-view is quite narrow imo.

core/ui/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/core/ui/date/relativeTime.kt
FelberMartin: Okay, then I thinks its fine :) 
Maybe we can add a `require()` statement at the top of the function so that it is clear that passing `showDateAndTime = true, showDate = false` would be considered illegal input.

Probably something like this:
``` 
val isShowDateParameterCombinationIllegal = showDateAndTime && !showDate
require(!isShowDateParameterCombinationIllegal)
```

FelberMartin (changes requested):
I am not sure, but I think the padding is still off, and now it looks like the profile picture is not aligned with the AuthorRole label anymore 🤔 
image

Regarding the issue on the foldable, I also noticed problems with this 3 column layout for another PR. Tbh I think we would be better off with just two columns when showing the ThreadUI (also on the tablet) with either 
(A) Make the ThreadUI replace the ChatList (this is how it currently is on iOS / iPadOS)
(B) Hide  the ConversationOverview on the left and display ChatList and ThreadUI side by side.

My take would be to do (A) for now (maybe in another PR?) and then think about what is the better strategy together with @eylulnc as soon as she deals with the navigation design as part of her thesis.

julian-wls:
> I am not sure, but I think the padding is still off, and now it looks like the profile picture is not aligned with the AuthorRole label anymore 🤔
I fixed it :)

> Regarding the issue on the foldable, I also noticed problems with this 3 column layout for another PR. Tbh I think we would be better off with just two columns when showing the ThreadUI (also on the tablet) with either (A) Make the ThreadUI replace the ChatList (this is how it currently is on iOS / iPadOS) (B) Hide the ConversationOverview on the left and display ChatList and ThreadUI side by side.
> 
> My take would be to do (A) for now (maybe in another PR?) and then think about what is the better strategy together with @eylulnc as soon as she deals with the navigation design as part of her thesis.

I agree. I think it would be good to do it as part of another PR to avoid chaos in this PR.

FelberMartin (changes requested):
Nice the alignment problem has been fixed now. However my original concern regarding the padding is still there, maybe I failed to explain it properly what my concern was. 

The padding from the post card to the inner card element and between the card elements is not consistent. See the screenshot below. The horizontal (red) and vertical (green) paddings of the card and the padding between the AuthorProfilePicture and the name to the right (blue) are all different values. In iOS all these paddings are of the same size and this looks much better imo.
image

FelberMartin:
I just noticed that in chats with just a few posts, the posts get displayed at the bottom now. Is this intentional?

image

julian-wls:
> Nice the alignment problem has been fixed now. However my original concern regarding the padding is still there, maybe I failed to explain it properly what my concern was.
> 
> The padding from the post card to the inner card element and between the card elements is not consistent. See the screenshot below. The horizontal (red) and vertical (green) paddings of the card and the padding between the AuthorProfilePicture and the name to the right (blue) are all different values. In iOS all these paddings are of the same size and this looks much better imo. !

Thanks for clarifying. I made the required changes and pushed them :)

> I just noticed that in chats with just a few posts, the posts get displayed at the bottom now. Is this intentional?

Yes, this is intentional to mimic the behavior of the iOS app.

FelberMartin:
The padding looks much better now, thank you for taking care of this :D

> Yes, this is intentional to mimic the behavior of the iOS app.
Hm okay, I just checked on the iPad version (because I have no iOS device), but there the posts are also displayed at the bottom. Can you verify this @julian-wls? Because also my intuition would be to display single posts at the top, like it is done eg in Whatsapp

julian-wls:
> The padding looks much better now, thank you for taking care of this :D
> 
> 
> 
> > Yes, this is intentional to mimic the behavior of the iOS app.
> 
> 
> 
> Hm okay, I just checked on the iPad version (because I have no iOS device), but there the posts are also displayed at the top. Can you verify this @julian-wls? Because also my intuition would be to display single posts at the top, like it is done eg in Whatsapp
> 
> 

I think it looks better on iOS. This is the behavior on iPhones:

image

FelberMartin (approved):
Interesting that it differs from iPhone to iPad, but then it's fine to keep it that way for Android 👍

eylulnc (approved):
Everything looks good.

PR #242, opened 2024-12-19
`Bugfix`: Fix crash when updating post with same clientPostId

### LINKED ISSUE #241: UniqueKeyConstraint Crash: Mark answer as resolving
### Describe the bug
I do not really have an idea what this is related to, so I will just describe how I could consistently reproduce the bug.

**To Reproduce**
- Login as test_user_1 on TS1
- Interactive Learning WS24/25 course
- Go to channel #random (I think this is required to load the message in this chat)
- Go back to Conversation Overview
- Open DM with test_user_20
- On the web app, as test_user_20
  - Go to same course and #random
  - Scroll up to post "testmessage" from Test User 16 Artemis with timestamp 2024-11-03
  - Show replies
  - Mark the second reply as resolving (or not resolving)
- Android app crashes

**Expected behavior**
No crash

### Screenshots
image

FelberMartin:
### Problem Description
The would crash when there were two posts with the same clientPostId in different conversations (which is allowed), and one received an update for post1 while being in the metisContext of post2. This happened due to a small bug in the implementation.

This closes #241 

### Changes
 
- Fixed bug
- Added test

### Steps for testing
Follow steps described in issue and see that the crash no longer happens:

> Steps to reproduce the behavior:
> - Login as test_user_1 on TS1
> - Interactive Learning WS24/25 course
> - Go to channel #random (I think this is required to load the message in this chat)
> - Go back to Conversation Overview
> - Open DM with test_user_20
> - On the web app, as test_user_20
>   - Go to same course and #random
>   - Scroll up to post "testmessage" from Test User 16 Artemis with timestamp 2024-11-03
>   - Show replies
>   - Mark the second reply as resolving (or not resolving)
> - Android app crashes

### DIFF
--- a/feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/service/storage/impl/MetisStorageServiceImpl.kt
+++ b/feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/service/storage/impl/MetisStorageServiceImpl.kt
@@ -148,13 +148,13 @@ internal class MetisStorageServiceImpl(
             serverId: String,
             clientSidePostId: String,
             serverSidePostId: Long?,
-            conversationId: Long = this.conversationId,
+            conversationId: Long? = null,
             postingType: BasePostingEntity.PostingType
         ): MetisPostContextEntity =
             MetisPostContextEntity(
                 serverId = serverId,
                 courseId = courseId,
-                conversationId = conversationId,
+                conversationId = conversationId ?: this.conversationId,
                 serverPostId = serverSidePostId,
                 clientPostId = clientSidePostId,
                 postingType = postingType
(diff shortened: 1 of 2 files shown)


julian-wls (approved):
Just tested it. Works as expected. Code lgtm! Just one question:
I think the changes in MetisStorageServiceImpl were introduced in #124. Did you check if the live insertion still works?

FelberMartin:
> I think the changes in MetisStorageServiceImpl were introduced in https://github.com/ls1intum/artemis-android/pull/124. Did you check if the live insertion still works?

Yep, just checked locally that the live insertion still works. The bugfix for this PR is actually just the change in line 504. The other three changes in the `MetisStorageServiceImpl` file were just refactoring, but the logic should be the same. Please correct me if I overlooked something 😅

julian-wls:
Okay sounds good :) 
Just wanted to make sure as I recalled changing something there a while ago. I also checked and there seems to be no issue with the live insertion, so I think it's fine 😁

PR #243, opened 2024-12-20
`Chore`: Koin upgrade followup

FelberMartin:
### Problem Description
There were still a few follow-ups after #135, these are tackled here.

### Changes
 
- Using new koin `verify()` method to verify module initialization
- Removing now no longer needed workaround to use koin in tests (`LocalKoinScope`, `LocalKoinApplication`)

### Steps for testing
- See that all tests still pass


julian-wls (approved):
Code lgtm!

PR #246, opened 2024-12-20
`Feature`: Change to Bottom navigation bar

FelberMartin:
### Problem Description
The Android app used a top tab bar for the top level navigation in the CourseUi. This is however not a modern approach and does not align with other Android communication app and the Artemis iOS version. On top of that (literally, lol), there was the problem that the course top bar was still displayed when navigating to a chat on the normal (phone) layout, which cluttered the screen and often lead to users clicking the wrong back button by accident.

### Changes
 
- Introduced a BottomNavigationBar
- Extracted the scaffold with the BottomNavigationBar and CourseTopBar, and only show it on the converationOverview or when in tablet layout.
- Chore:
  - Extracted setup of NavGraph into separate files
  - Removed unused navigation code
  - Using `NavigationBackButton` everywhere

### Steps for testing
On both phone and tablet check that the navigation still works as expected, eg:
- Navigating between the (Conversation) screens
- Using deeplinks (eg via user mentions in a post, or savedPosts feature)
- BottomNavBar is displayed only for the ConversationOverview on Phone layout
- CourseTopAppBar is hidden on Phone layout when in a chat

### Screenshots

https://github.com/user-attachments/assets/aed1c2a2-b48a-46cd-a857-b57df9113a24


eylulnc (changes requested):
This is looking really good! I am always a bottom navigation person so 😍.
I have few small points.
1 - In the tablet view test area is really close to the bottom-nav-menu. Maybe padding can be add?
Screenshot 2025-01-25 at 12 17 47

2 - In the conversation disabled case, We navigate and saw in very dark background, communication disabled message.
Screenshot 2025-01-25 at 12 21 51

In iOS they just do not show the tab, but I prefer to see the tab instead of having to only 2 tab on the bottom, maybe we can stay in the same navigation level and background so that message is more visible as in the early version. What do you think?

Screenshot 2025-01-25 at 12 24 26

FelberMartin:
>  1 - In the tablet view test area is really close to the bottom-nav-menu. Maybe padding can be add?

I will add a small padding for now, but I assume we would probably have to undo the padding in Julians ReplyTextField refactoring PR #320. I will keep in mind to check this before merging the later of these two PRs :)

> 2 - In the conversation disabled case, We navigate and saw in very dark background, communication disabled message.

Good catch, this was just because I missed to use the scaffold in the case where communication is disabled. The CommunicatinDisabledInfo now shows as in the previous versions again :D 

image

FelberMartin:
FYI: Due to the recent merge of #320, the padding of the ReplyTextField is off now. I will have a look today

julian-wls (commented):
Really nice change, this makes navigation a lot better!

Only thing I noticed while testing is that the BasicDataStateUi is not visible in the ConversationOverview when using pull to refresh, instead the screen becomes black:

https://github.com/user-attachments/assets/2aa725eb-5695-4085-825c-e8d6ca58af4e

Couldn't find issues in the code.

PR #252, opened 2024-12-21
`Bugfix`: Only show resolve post if actionable

### LINKED ISSUE #240: Resolve post action shows for other people posts
### Describe the bug
In eg a DM or group chat, all users see the "Resolves Post" action when long clicking a thread answer. However clicking this has no effect if the user is not the post author.

**To Reproduce**
I think this should work on any server, course and chat, but here is my setup:
- Login as test_user_1 on TS1
- Interactive Learning WS24/25
- DM chat with test_user_20
- Click test_user_20's post with replies (or create a new reply)
- Long press on the reply
- "Resolves Post" action shows
- Pressing the action does not mark the thread as resolved

**Expected behavior**
Either pressing the resolve button should resolve the post properly, or the option should not show up in the Actions sheet at all.

### Additional context
Investigate how the behaviour is on the webapp (can users mark other user's posts as resolved?) and adjust either the action visibility on Android or fix that clicking "Resolves Post" actually does what its supposed to do.
Maybe it is also important what role the current user has. Eg on the webapp I could observe that test_user_20 (instructor) is able to resolve test_user_1's post in a DM chat.

FelberMartin:
### Problem Description
The "Resolves Post" action did wrongly show up for when selecting a users own answer post, even when the user has no tutor rights and he is not the thread base post author.

This closes #240.

### Changes
 
- Fixed the bug
  - Changed `IAnswerPost` to include the parentAuthorId
  - Adapted inheriting classes
- Changed ThreadUI to make use of `IStandalonePost` instead of `PostPojo` for easier testing
- Created tests
- Restructured test file

### Steps for testing
Prerequisites: user without tutor rights
- Go to a other user's post
- Write a new answer to this post
- Long press on your answer
- The "Resolves Post" action does not show anymore


eylulnc (approved):
Code looks good to me.
Also implementation too, I have check it for both student and instructor.
Resolve not visible for students ✅

julian-wls (approved):
Just tested it and the resolve post action does not show up.
Code lgtm, nice cleanup of the test file!

PR #253, opened 2024-12-22
`Feature`: Add Unordered and Ordered List Options to Formats

eylulnc:
### Problem Description
Our mobile markdown editor lacked support for creating ordered or unordered lists, a feature already available in the web editor. Mobile users were unable to format lists conveniently, requiring them to manually type prefixes like -  or 1. , which disrupted the ease of composing structured messages.

### Changes
 
**Ordered List:** Adds logic to insert a 1.  prefix (or increment an existing number, e.g. 2.  -> 3. ) when the user selects “Ordered List.”

**Unordered List:** Adds logic to insert a -  prefix.

**Formatting Dropdown:** Introduces a single “List” icon in the toolbar. Tapping it reveals a dropdown to choose between unordered/ordered lists.

### Steps for testing
1.	Open any conversation
2.	Tap the “List” icon. Verify a dropdown appears with “Unordered List” and “Ordered List.”
3.	Choose “Unordered List” to insert or apply -  at the cursor.
4.	Choose “Ordered List” to insert or apply 1.  (or increment an existing number).
6.	Check auto-continuation: If a line starts with -  or N. , pressing Enter automatically inserts another bullet or the next number.

### Screenshots

https://github.com/user-attachments/assets/8a8cc416-2607-4588-93d5-1fb6d637204b

Screenshot 2024-12-22 at 21 31 05


FelberMartin (changes requested):
Very nice, I like this enhancement!

I have a few minor notes I noticed during testing:
1. When clicking the list icon, the keyboard is collapsed. Would be nice if we could prevent this from happening, but tbh I am not sure whether / how this is possible.
2. From the implementation of the `continueList` function I take that the list autocompletion should also work when the cursor is not at the end of the text. But during testing, this scenario did not work. I think this may be related to the check in line 823. 
3. This is not directly related to this PRs scope, but if it is not too much effort, we could also include this here: When I switch from "Edit" to "Preview" mode, the formatting options are still shown. I think these should be hidden in this context. But if there is no easy fix, we can create a separate issue for this :D 

ArtemisMarkdownListAutocompletion

feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/reply/ReplyTextField.kt
```diff
@@ -694,3 +764,118 @@ private fun String.takeWhileTag(): String {
...
+/**
+ * If the previous line starts with "-", continue as unordered.
+ * If it starts with "N. ", continue as ordered.
+ * Otherwise, do nothing (return the original newValue).
+ */
+private fun continueListIfApplicable(
+    oldText: String,
+    newValue: TextFieldValue
+): TextFieldValue {
+    val afterUnordered = continueUnorderedList(oldText, newValue)
```
FelberMartin, line 778: I'm thinking, maybe we should move this whole list continuation logic out of this file into a `MarkdownListContinuationUtil` file. This would allow us to write some basic unit tests to ensure the proper behaviour and also avoid bloating the `ReplyTextField.kt` file

FelberMartin (changes requested):
Nice, points 2. and 3. of my previous review are fixed now :D

Did you take a look at 1.? When I checked this issue persisted

feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/reply/ReplyTextField.kt
eylulnc: I will take a look at this and for the !.st point I could not find any solution

feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/reply/MarkdownListContinuationUtil.kt
```diff
@@ -0,0 +1,120 @@
...
+import androidx.compose.ui.text.TextRange
+import androidx.compose.ui.text.input.TextFieldValue
+
+
+/**
+ * If the previous line starts with "-", continue as unordered.
+ * If it starts with "N. ", continue as ordered.
+ * Otherwise, do nothing (return the original newValue).
+ */
+fun continueListIfApplicable(
```
FelberMartin, line 12: Please wrap the methods in this file in a `object MarkdownListcontinuationUtil`, so that the call to the util functions is more explicit. I think you can also mark all methods besides the called one as private

feature/metis/conversation/src/test/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/reply/MarkdownListContinuationUtilTest.kt
```diff
@@ -0,0 +1,77 @@
...
+            selection = TextRange(cursor, cursor)
+        )
+
+        val result = continueListIfApplicable(oldText, newValue)
+
+        val expectedText = "Hello\n"
+        assertEquals(expectedText, result.text)
+        assertEquals(expectedText.length, result.selection.start)
+        assertEquals(expectedText.length, result.selection.end)
+    }
```
FelberMartin, line 74: Please add a test to cover list continuation when the cursor is not at the end of the text (to cover the scenario of point 2. of my previous change request)

feature/metis/conversation/src/test/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/reply/MarkdownListContinuationUtilTest.kt
```diff
@@ -0,0 +1,77 @@
...
+import androidx.compose.ui.text.input.TextFieldValue
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import de.tum.informatics.www1.artemis.native_app.core.common.test.UnitTest
+import org.junit.Assert.assertEquals
+import org.junit.Test
+import org.junit.experimental.categories.Category
+import org.junit.runner.RunWith
+
+@Category(UnitTest::class)
+@RunWith(AndroidJUnit4::class)
```
FelberMartin, line 13: Use `@RunWith(RobolectricTestRunner::class)` instead, as we do not need Android specific behaviour for this test.

feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/reply/ReplyTextField.kt
```diff
@@ -269,7 +277,15 @@ private fun CreateReplyUi(
                     },
                     onFileSelected = { uri ->
                         onFileSelected(uri)
-                    }
+                    },
+                    formattingOptionButtons = {
+                        if (displayTextField || currentTextFieldValue.text.isNotBlank()) {
```
FelberMartin, line 282: Did you check whether this if statement is required?

FelberMartin (changes requested):
Some minor changes :)

feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/reply/ReplyTextField.kt
eylulnc: I believe this was supposed to make the formatting options visible only in edit mode, but it didn’t work as expected. I removed it and tested again but didn’t notice any difference in behavior.

eylulnc:
> Nice, points 2. and 3. of my previous review are fixed now :D
> 
> Did you take a look at 1.? When I checked this issue persisted

1. should be fixed. Can you verify?

FelberMartin (approved):
Can confirm, pressing the listing formatting option does no longer collapse the keyboard, this is awesome :D

Ready to merge from my side

julian-wls (approved):
Tested it and works as expected.
Code lgtm.

PR #257, opened 2024-12-27
`Chore`: Set tum artemis server as default

FelberMartin:
### Problem Description
The current Artemis Android release in the Play Store uses the "Tum" build flavor, which does not allow instance (artemis server) selection. Simply switching the released build flavor to "unrestricted" would have caused already logged in users with weird behaviour (all network calls would fail because it would try to reach the artemis server via localhost; only a logout and sign in again would have resolved this issue).

### Changes
 
- Changed the default server for the "unrestricted" flavor to the tum artemis server
- Adapted settings code to rely on the `serverUrl` instead of the duplicated logic with `hasUserSelectedInstance`
  - Removed now no longer needed code `hasUserSelectedInstance`
- Fixed bug introduced by https://github.com/ls1intum/artemis-android/pull/222, as it removed the `NestedDestination.InstanceSelection` navigation target, but did not remove it in other places of the code. This would lead to an app crash just after freshly installing the app.

#### Scenario 1 - Updating the app from "tum" to "unrestricted" flavor
- Uninstall the app on the test device
- Switch the build variant by Menu > Build > Select Build variant ... and select "productionTumDebug" for `:app`
- Launch the app and log in with your TUM credentials
- Now we check that changing to new release version with the "unrestricted" flavor works as expected:
  - Change the build variant to "productionUnrestrictedDebug" for `:app`
  - Launch the app
  - See that you are still logged in with your account to the TUM server and the dashboard is loaded

#### Scenario 2 - Fresh install
- Uninstall the app on the test device
- Directly change the build variant to "productionUnrestrictedDebug" for `:app`
- Launch the app
- See that by default the tum instance is selected for login, but the user can also select a different server by pressing "not your university"

### DIFF
--- a/build-logic/convention/src/main/kotlin/commonConfiguration/KotlinAndroid.kt
+++ b/build-logic/convention/src/main/kotlin/commonConfiguration/KotlinAndroid.kt
@@ -134,6 +134,8 @@ internal fun Project.configureReleaseTypeFlavors(
     }
 }
 
+private const val TUM_ARTEMIS_SERVER_URL = "https://artemis.cit.tum.de"
+
 internal fun Project.configureInstanceSelectionFlavors(
     commonExtension: CommonExtension<*, *, *, *, *, *>,
 ) {
@@ -153,7 +155,7 @@ internal fun Project.configureInstanceSelectionFlavors(
                 buildConfigField(
                     "String",
                     ProductFlavors.BuildConfigFields.DefaultServerUrl,
-                    "\"\""
+                    "\"$TUM_ARTEMIS_SERVER_URL\""
                 )
             }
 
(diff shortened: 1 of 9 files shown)


eylulnc:
Both scenario works as expected, I only have few suggestions,

1 - Maybe we can open the instance bottom sheet as full
2 - In iOS if we select different instance that default,  itshows which one maybe we can also have sth like this

Screenshot 2024-12-30 at 12 28 25

eylulnc (approved):
Lgtm!

FelberMartin:
I addressed your comments and the BottomSheet now opens in full screen. 
I also added an indicator of the the server host, however I thought it mainly is helpful for us developers (which test server do we have selected?), but not really necessary for the end users, so I decided to only show it in the debug version. Do you think this is sufficient @eylulnc ?

julian-wls (approved):
Just tested your changes and I did not notice any issues. The code changes seem appropriate.

PR #272, opened 2025-01-06
`Chore`: Fix paddings

FelberMartin:
### Problem Description
The usage of padding has not been consistent over the app and in some places used wrongly. This resulted in some UI element not being visible, especially on the tablet layout with a larger NavigationBar (eg selecting a different server instance, or the button for accepting the code of conduct)

### Changes
 
- Correctly consume window insets with `consumeWindowInsets()` (often only top padding was applied but the hole `WindowInsets.systemBars` was consumed, which also implies the `WindowInset.navigationBars` insets at the bottom)
- Usage of common spacing values from `Spacings` where applicable
- Minor refactorings along the way

### Steps for testing
Look through the screens of the app. Nothing should have changed visually (apart from InstanceSelection etc being visible now on the tablet)


julian-wls (commented):
Code lgtm, just one question:

When going through the app I noticed that now there is no visual separation between the bottom sheets and the status bar (See screenshot). Is this intended? I don't think it's bad, I just think it would be nice to have a little bit of background behind the bottom sheets and not open them full-screen.

Screenshot:
ScreenShot2

FelberMartin:
Yep, previously this behaviour was inconsistent between the different BottomSheets, so I aligned all of them to show behind the status bar. However I don't have a strong opinion about which option is better. @eylulnc do you have a preference?

julian-wls (approved):
Looks better now!

Issue #265, opened 2025-01-03
Pull to refresh instead of reload button


julian-wls:
### Description
Currently, there are a lot of reload buttons in all top bars, which seem a little bit outdated.

### Solution
In order to give the app a better look it would be good to replace the reload button with a pull to refresh mechanism similar to the iOS app, see [official docs](https://developer.android.com/reference/kotlin/androidx/compose/material/pullrefresh/package-summary).
This mechanism could be added to all lazylists to replace the reload button.

PR #276, opened 2025-01-07
Replace Refresh Buttons with Pull to Refresh & change Top Bar scroll behavior

julian-wls:
### Problem Description

As described in #265 the refresh buttons in the top bar seem quite outdated.

### Changes

This PR introduces Pull to refresh by changing the box around DataStateUi to a PullToRefreshBox, for the mechanism to work correctly the top bar has been changed to a non collapsible Top Bar. Pull to refresh is now available in all views, that use a dataStateUi and have pull to refresh enabled. 
The refresh buttons have been removed and the refresh buttons in the chat list and in the thread view are only available for debugging now.
Placeholder labels have been removed in the top bar as they were visible when triggering the reload. 

Note: The exercise view still uses a refresh button, as this view is currently not scrollable. This will be changed in a later PR, that cleans up the exercise view.

### Steps for testing
1. Go through different views of the app
2. Pull to refresh and make sure it works correctly.
3. Verify that the behavior of the Top bar still works as expected.


FelberMartin (approved):
Awesome change, this makes the app feel so much better! :D

One thing I noticed is that pull to refresh is available in the thread view, but when using it, the screen is changed back to the ConversationOverview. I think this may be related to how navigation is currently handled and I would not be surprised if there is no straight-forward fix for it. However I think it's still fine if we keep it as is, but just look whether there is an easy fix :D

julian-wls:
> One thing I noticed is that pull to refresh is available in the thread view, but when using it, the screen is changed back to the ConversationOverview. I think this may be related to how navigation is currently handled and I would not be surprised if there is no straight-forward fix for it. However, I think it's still fine if we keep it as is, but just look whether there is an easy fix :D

Yes, that's correct. I noticed this behavior, too. The problem is caused by layering multiple DataStateUi's over each other. CourseUi has one, Code of Conduct has one and Conversation overview has one. I disabled pull to refresh for the subviews, as this would otherwise cause a conflict. Unfortunately, we cannot disable pull to refresh in CourseUi in favor of the communication sub-views, as the exercise tab and the lecture tab do not use any sub-DataStateUi's. I haven't found a solution to this problem yet. @FelberMartin what do you suggest?

FelberMartin:
Okay then I'd say lets keep it as is, create a follow-up issue, and then we can maybe discuss in more detail next Tuesday before/after or weekly meeting :)

PR #279, opened 2025-01-09
`Chore`: Replace three column with two column conversation layout for large screens

FelberMartin:
### Problem Description
On tablets the layout with the ConversationOverview, ChatList and ThreadUI all displayed at once is quite packed and results in many UI elements being displayed incorrectly, due to the little space available. Additionally the iOS tablet version also does not support the 3-column layout. So by changing to a two column layout we also align the mobile experiences.

### Changes
 
Changed the ConversationScreen to only show two columns at most. 

### Steps for testing
On tablet or foldable:
- Go to ConversationOverview
- Open Chat -> Chat is displayed on the right
- Navigate to thread -> ConversationOverview on the left + Thread on the right is displayed.

On smartphone:
- Everthing works as before

### Screenshots

https://github.com/user-attachments/assets/cc308099-b661-4dfb-a06d-d313f5e551ca


julian-wls (approved):
Tested it on tablet, works as expected and looks a lot better. Thanks for taking care of this :)
Code lgtm!

Issue #274, opened 2025-01-06
Chats not loading on tablet?


FelberMartin:
### Describe the bug
On the two column layout there seems to be an issue with loading the chats when switching the via the conversation overview on the left. Judging from the network inspector, the network calls simply receive no answer.

**To Reproduce**
Steps to reproduce the behavior with a tablet:
1. Go to the ConversationOverview
2. Select a chat -> wait until its loaded
3. Select a different chat from the ConversationOverview on the left
4. Notice that there is a infinite loading of the chat

**Expected behavior**
The selected chat should be loaded.

### Screenshots

https://github.com/user-attachments/assets/c919ffab-3ebc-4f49-9f15-a23490b64aa4

### Additional context
Interestingly this only happens with the two columns layout (so not on the smartphone). 

Also the chat is loaded again when navigation back to the ConversationOverview-only view (by clicking the back button on the chatList).

FelberMartin:
Can someone confirm that this issue is not unique to me? I could only check on the emulator

Issue #267, opened 2025-01-03
Back Navigation in ThreadUI not working


FelberMartin:
### Describe the bug
When the user is in the ThreadUI screen and presses the Android back button (not the BackArrow in the top left), the user stays in the Thread screen. After pressing the back button a second time, the user is navigated back to the conversation overview, skipping the ChatList screen.

**To Reproduce**
1. Go to any Chat
2. Click on a post to open the thread screen
3. Press the Android back button -> nothing happens
4. Press the back button again -> the ConversationOverview is shown

**Expected behavior**
Pressing the back button once should show the ChatList screen.

PR #280, opened 2025-01-09
`Bugfix`: Chats not loading on tablet layout

FelberMartin:
### Problem Description
This PR addresses multiple bugs / issues:
1. #274 
2. #267 
3. When loading a chat that has not been opened before, the ChatList shortly displayed "No posts found" before showing the loaded posts.

### Changes
 

- To fix 1: Removed the `viewModelStoreOwner` introduced by Tim Ortel in https://github.com/ls1intum/artemis-android/pull/30. I talked to him, but he could not remember why exactly this approach with the store was introduced in the first place. I could not find any bugs related to my removal of this code, I tested both smartphone and tablet layouts, navigation between chats and threads, disconnecting from the internet, etc. 
- To fix 2: Update `showThread` properly
- To fix 3: Check for `loadStates.isIdle`
- Misc: Reset scroll position when entering new chat (needed when switching chats in tablet layout)

### Steps for testing
On both tablet and smartphone test:
- Go to a chat
- Go to another chat (on tablet via the ConversationOverview on the left) -> chat is loaded properly
- Open a thread
- Close a thread both via the Android back button and the arrow icon in the TopBar -> both work
- On tablet: Open a thread and select a different chat on the left -> new chat is shown in the right column

Please also thoroughly test the changes to fix bug 1., and test any scenario that could be affected by removing the viewModelStoreOwner.


FelberMartin:
### For future reference
In case we discover any problem related to the removal of the  `viewModelStoreOwner`, here is a code snippet that also fixes Bug 1:

``` 
    val lifecycle = LocalLifecycleOwner.current.lifecycle
    val lifecycleObserver = rememberUpdatedState(
        LifecycleEventObserver { _, event ->
            if (event == Lifecycle.Event.ON_DESTROY) {
                store.clear()
            }
        }
    )

    DisposableEffect(lifecycle) {
        val observer = lifecycleObserver.value
        lifecycle.addObserver(observer)
        onDispose {
            lifecycle.removeObserver(observer)
        }
    }
```

I decided to not insert this code right away, because we currently do not know why the code is even needed. If we discover the reason, please also insert a detailed comment explaining why this code is needed.

eylulnc (approved):
I think it looks good.
Tested in both phone, tablet.
Chats are visible. ✅
Navigating back from chat leads to conversation list. ✅
No chat found is not display if a content exist. ✅

julian-wls (approved):
Great work!
I test all scenarios on both, tablet and phone.
The Android back button is also working now leading the user back to the conversation list.
Your changes seem appropriate and can be merged imo.

Regarding the removal of `viewModelStoreOwner` I couldn't find any issues or bugs. I don't find any reason why `viewModelStoreOwner` should be used here either.

PR #292, opened 2025-01-11
`Improvement`: Improve Thread header and UnfocusedReplyTextField

FelberMartin:
### Problem Description
1. In conversations with long names (eg in GroupChats with the default group name), the UnfocusedReplyTextField could span multiple lines, which does not look consistent.
2. The ThreadView TopBar title was always "View Post", which is quite non-informative. This is replaced with a more informative header, inspired by Artemis iOS and Slack's design.

### Changes
 
- Applied `maxLines = 1` to conversationNames
- Introduced the conversation context to the thread header
- Fixed a problem with ProfilePictures displaying at small sizes
- Removed unused string resources

### Steps for testing
- Go to a chat with a long name -> notice that the ReplyTextField is only a single line and is truncated
- Go to any thread -> notice the changes in the header

### Screenshots
1 Previously
grafik

1 After 
grafik

2 Previously
grafik
2 After
grafik


julian-wls (approved):
Really nice changes, works as expected!
Code lgtm!

Note: I created a [follow up PR](https://github.com/ls1intum/artemis-android/pull/297) to also show the number of members in the chat.

Issue #293, opened 2025-01-11
Show profile picture in notifications


FelberMartin:
### Description

The communication notifications currently only show a generic chat icon for incoming notifications. This should be replaced with the post authors profile picture.

This is a sub-issue of #216

Issue #216, opened 2024-12-13
Notification improvements


FelberMartin:
There are currently a few improvement ideas for enhancing the notifications or fixing bugs:

- ✅ The notifications currently only show a generic profile picture (probably created by the authorName) and does not use the Artemis specific profile pictures
  - Implementation hints: In `CommunicationNotificationManagerImpl.buildMessagingStyle` we would need to use `person.setIcon(...)` and previously store the image url in the `CommunicationMessageEntity`
  - Update: Has been extracted to #293 
- Reading a message does not automatically remove the notification (like eg in WhatsApp) -> Blocked by #628 
- ✅ When pressing "Mark as read" on a notification, the message still shows as unread when later entering the app
  - Update: Done in https://github.com/ls1intum/artemis-android/pull/481
- When using the quick reply feature of a notification, show the users reply in the notification (see `UpdateReplyNotificationWorker.kt`)

PR #294, opened 2025-01-11
`Chore`: Refactor notifications code

FelberMartin:
### Problem Description
As I was looking into the notifications code to prepare for #293, I found multiple opportunities for refactoring and improving the readability of the code. For a cleaner PR scope I decided to already open this PR as a pure refactoring PR and followup with the actual PR separately.

### Changes
 
- Refactoring of `handleServerPushNotification()`
- Removed no longer used "title" property in notifications code
- Introduced `CommunicationNotificationPlaceholderContent` for easier notification parsing

### Steps for testing
Test that receiving notifications still works.
Known issues are listed in #216


feature/push/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/push/communication_notification_model/CommunicationNotificationPlaceholderContent.kt
```diff
@@ -0,0 +1,119 @@
...
+        val answerPostContent: String,
+        val answerPostAuthor: String,
+    ) : CommunicationNotificationPlaceholderContent() {
+
+        companion object : NotificationPlaceholderContentFactory {
+            override fun fromNotificationPlaceholders(notificationPlaceholders: List<String>): CommunicationNotificationPlaceholderContent {
+                return ReplyPost(
+                    courseTitle = notificationPlaceholders[0],
+                    postContent = notificationPlaceholders[1],
+                    postAuthor = notificationPlaceholders[3],
```
julian-wls, line 111: Why is the content of `notificationPlaceholders[2]` not needed here?

feature/push/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/push/service/impl/PushNotificationHandlerImpl.kt
```diff
@@ -35,6 +35,22 @@ class PushNotificationHandlerImpl(
     }
 
     override fun handleServerPushNotification(payload: String) {
+        if (!validatePushNotificationVersion(payload)) return
+
+        val notification = decodeNotification(payload) ?: return
+
+        // if the metis context this notification is about is already visible we do not pop that notification.
```
julian-wls, line 42: What do you mean here :)

julian-wls (approved):
Receiving notifications still works as before, I didn't experience any issues while testing.
Changes seem appropriate.
Just two comments:

feature/push/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/push/service/impl/PushNotificationHandlerImpl.kt
FelberMartin: This was a leftover from the previous, less structured implementation. It tries to explain what `        if (isNotificationContextAlreadyObservedByUser(notification)) return
` does, but I think after the refactoring this is probably clear from the method name, so maybe its fine if I just remove the comment. What do you think?

FelberMartin:
After talking with Anian this week, I decided to more closely match the iOS implementation (as it is also more up-to-date). Could you therefore have a look again? @julian-wls

julian-wls (approved):
Functionality is the same as before, didn't experience any issues. 
The code looks indeed more up-to-date and more structured.
Re-approve.

PR #297, opened 2025-01-15
`Improvement`: Add member count to chat top bar title

julian-wls:
### Problem Description
The member count is not shown in the Top bar like it is shown on iOS.

### Changes
The Top Bar title of the chat has been changed to a column to also show the number of members if the conversation is not a DM chat.

### Steps for testing
1. Verify that the member count is represented correctly in a chat
2. Verify that it is not shown for DM chats

### Screenshots
xxx


FelberMartin (approved):
Nice change!

One question:
Was it an active decision to then also use the smaller title font size for DM chats? Tbh I dont know whats better design-wise, just wondering whether you considered this :D

julian-wls:
> One question: Was it an active decision to then also use the smaller title font size for DM chats? Tbh I dont know whats better design-wise, just wondering whether you considered this :D

I figured it would be best to stay consistent with the font here, that's why I also kept the same style for the DM chats.
On iOS it's also done like this.

PR #308, opened 2025-01-16
`Improvement`: Markdown code block adjustments

julian-wls:
### Problem Description
Code blocks in markdown text look different from the iOS and web app.

### Changes

This PR changes the colors of the code blocks and adds rounded corners.

### Steps for testing
Make sure the code block still looks good.

### Further investigation

- Note: Outside the scope of this PR, but worth discussing: I found a way to add a corner radius to images as well, but `wrapContentSize` is needed to make it work properly. Unfortunately, this caused some of the images in the chat to be too small. The approach was similar to clipping code blocks. Does anyone have an idea how to make this work? I would be happy to discuss this somewhere. 
- I found out, that there is a library available, that could enable syntax highlighting to code blocks. Would it be worth taking a look?


FelberMartin (approved):
Looks generally good to me

FelberMartin:
I noticed that the rounded corners get applied to the whole posts that start with a code block. That leads to some inconsistencies, eg when there is a code block in the middle of a post, or when there is eg a trailing image.

First code block inside a post, the second one is in a standalone post. Notice the different corners:
image

Post starting with code block but with trailing image. Notice the inconsistent corners on the image (only bottom left corner is rounded)
image

julian-wls:
I just noticed this, too. Let's put this PR on hold for now and discuss the details on Thursday. I think with a similar approach we can manage to apply the rounded corners correctly, maybe for images, too.

FelberMartin:
@julian-wls What is the status on this PR? In case you do not plan to work on this in the next week, I'd suggest to maybe close this PR for now and re-open it later, so we can keep the PR overview page clean 😅

julian-wls:
> @julian-wls What is the status on this PR? In case you do not plan to work on this in the next week, I'd suggest to maybe close this PR for now and re-open it later, so we can keep the PR overview page clean 😅

Yes sure :) I'm going to revert the rounded corners for code blocks as this didn't lead anywhere so we only have the color adjustments and can put it back in review.

FelberMartin (approved):
Just one code comment, good to merge from my side :)

core/ui/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/core/ui/markdown/MarkdownRenderFactory.kt
```diff
@@ -107,9 +107,14 @@ object MarkdownRenderFactory {
                 builder
                     .linkColor(context.getColor(R.color.link_color))
                     .isLinkUnderlined(false)
+                    .codeTextColor(context.getColor(R.color.code_text_color))
```
FelberMartin, line 110: The enclosing function and value name could probably be changed to `createTextThemingPlugin()` or similar

PR #311, opened 2025-01-17
`Chore`: Make use of isRequesting user instead of passing down clientId

FelberMartin:
### Problem Description
I noticed that there is a field called `isRequestingUser` when requesting conversationUsers that is true for the app user. This way we can easily identify the partner in a oneToOneChat, instead of having to pass down the app users clientId.

### Changes
 
- Making use of the isRequestingUser value
- Adjust tests

### Steps for testing
Ensure that the following things still work as expected:
- Display profile picture in DM chat header
- Navigate to DM chat by clicking on user mention in post


julian-wls (approved):
Profile picture in DM chat header is still displayed as expected,
Navigate to DM chat works as expected.

Code lgtm!

PR #316, opened 2025-01-18
`Bugfix`: Fix profile picture flickering when sending post

### LINKED ISSUE #203: Post items get recomposed when new post is added
### Describe the bug
Post items get recomposed when a new post is added to the chat, causing the profile pictures to reload. This happens in both the main ChatList and the ThreadUi.

**To Reproduce**
- Go to chat with at least one post of a user with profile picture.
- Send a new message
- See that the profile picture quickly disapears

**Expected behavior**
The profile picture should stay there and not flicker.

### Screenshots
[If applicable, add screenshots to help explain your problem.](https://github.com/user-attachments/assets/91928886-4505-4afa-9a0e-4d0243fbcbd9)

### Additional context
It seems that the root cause of this is that the post items get recomposed when a new post is added. This did previously not cause a visual problem, as the text is redrawn instantly, but with the profile pictures this is noticeable. 

Maybe helpful: https://stackoverflow.com/questions/74699081/jetpack-compose-lazy-column-all-items-recomposes-when-a-single-item-update

FelberMartin:
### Problem Description
After the introduction of the profile pictures in posts, there has been a bug where the profile picture would flicker when the user sends a new message (post or answer). Originally I thought this bug was related to the posts being falsely recomposed, so I looked into the recomposition of the posts and LazyColumns. While I found some optimizations on the way, the core issue turned out to not being related to recomposition at all (see changes section).

This closes #203 

### Changes
 
- Bugfix: pass the author's imageUrl when first locally creating the post before sending it to the server.
- Optimizations done on the way:
  - Use the default coil memoryCacheKey instead of defining it manually
  - Mark post interfaces as `@Stable` for compose optimizations. (All implementations of these interfaces are data classes and only used in a stable manner). This allows compose to skip recompositions more often
  - Moved the notification payload examples to a different directory as there where compilation problems

### Steps for testing
User an account with a profile picture, do:
- Go to any chat (or thread)
- Send a new message
- -> Notice that the profile picture no longer flickers 🥳


julian-wls (approved):
Really nice change, good catch!
Just tested it and the flickering seems to be gone

PR #318, opened 2025-01-20
`Improvement`: Increase space below notification settings categories

julian-wls:
### Problem Description

When scrolling to the end of the notification settings, the floating action button to save the changes is placed right above the switch of the last notification category making it hard for users to reach this switch.

### Changes

A spacer has been added below the list to increase the padding after the last category. This allows users to reach the switch more easily.

### Steps for testing

Scroll to the end of the notification settings and verify that there is enough space now to reach both, the last switch and the floating action button.

### Screenshots
Screenshot5


FelberMartin (approved):
Changes lgtm

PR #330, opened 2025-01-26
`Feature`: Mark All Channels as Read

eylulnc:
### Problem Description
Currently, users must manually go through conversations one by one to mark messages as read. This process is inefficient, and there is no functionality to mark all conversations as read at once.

### Changes
 
Added a "Mark All as Read" button to the existing FAB menu.
On Failure: Displays a pop-up dialog informing the user of the failure.

### Steps for testing
1. Navigate to the **Conversation Overview** that has unread messages.
2. Click the FAB to open the menu.
3. Verify that the "Mark All as Read" button appears in the FAB menu.
4. Test the "Mark All as Read" button:
    - Success Scenario: All conversations with unread messages are marked as read (unread counts reset to 0).
    -  Failure Scenario: A dialog pops up with an appropriate error message.

### Screenshots

Screenshot 2025-01-26 at 20 25 24
Screenshot 2025-01-26 at 20 30 09


feature/metis/manage-conversations/src/test/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/manageconversations/overview/ConversationOverviewE2eTest.kt
```diff
@@ -222,6 +223,24 @@ class ConversationOverviewE2eTest : ConversationBaseTest() {
...
+        assertEquals(3L, chat.unreadMessagesCount)
+
+        conversationService.markAllConversationsAsRead(
+            courseId = course.id!!,
+            authToken = accessToken,
+            serverUrl = testServerUrl
+        )
+
+        chat = chat.withUnreadMessagesCount(0) as OneToOneChat
+        assertEquals(0L, chat.unreadMessagesCount)
```
FelberMartin, line 241: This test does actually test the `withUnreadMessagesCount` method rather than the `markAllConversationsAsRead` endpoint. When you comment out the service call, the test still passes.

I think we would have two possibilities:
1. After creating the chat, send a new post from a different account to increase the unread messages count, then all the service endpoint, and then fetch the chat again and assert that the unread messages count is 0.
2. As the first approach I not straight forward and we would actually just test that the server behaviour is correct (which the server tests should already do), I think it might be fine to just test whether a call to the API returns true, so that the test fails in case the api changes. 

Would do you guys think?

eylulnc: I think I will go with the second option and only validate if the api returns success

FelberMartin (approved):

PR #341, opened 2025-01-26
`Improvement`:  Support monochrome app icon

FelberMartin:
### Problem Description
In newer Android versions (starting from Android 13 (API level 33)), the user can enabled themed (monochrome) icons on their home screen. The Artemis Android app did not yet support this.

### Changes
 
Regenerated ic_launcher icons and supporting monochrome app icon.

### Screenshots
Previously:
image

Now:
image


julian-wls (approved):
Nice change, good catch!

PR #343, opened 2025-01-27
`Chore`: Add type safe deeplinks

FelberMartin:
### Problem Description
The path definition of deeplinks was not defined in a central place, but instead used all over the app's code. With a more structured approach using deeplinks becomes much easier and one can easily get a glimpse of where they are used.

### Changes
 
- Introduced central `ArtemisDeeplink` class
- Implemented feature separated sub-classes of `ArtemisDeeplink`
- Used these classes where possible in the app
- Chore: removed some unused notification code

### Steps for testing
Navigate through the app using deeplinks (eg saved posts, posts with links to lectures/exercises/chats)


eylulnc (approved):
Tested links via save post and channel labels. Working correctly.

julian-wls (approved):
Really nice changes, thanks for taking care of this.
Tested it and work as expected.
Code lgtm.

PR #344, opened 2025-01-27
`Chore`: Migrate remaining colors to new color organization

### LINKED ISSUE #48: Global color organization
### Problem
Colors are not quite organized in a structured way, sometimes colors are hardcoded, sometimes they are only defined for light-mode and often they are defined redundantly.
### Suggestion
Handling the colors using a global color palette could be a great improvement to create a consistent way to deal with multiple colors used in various UI elements. The [iOS Artemis App Color Palette](https://github.com/ls1intum/artemis-ios-core-modules/blob/main/Sources/DesignLibrary/Colors.swift) could serve as an inspiration to implement something similar.

For more information see the discussion [here](https://github.com/ls1intum/artemis-android/pull/43#discussion_r1787056452).

julian-wls:
### Problem Description

The colors used across various places in the app where not defined in a consistent way. Over the time we adopted a new way to organize colors in the app by using objects.

### Changes

This PR migrates the remaining colors that were not yet part of the new organizational approach.
This PR closes #48.

### Steps for testing

1. Verify that there a no left overs and make sure that all colors are organized in a structured way.


core/ui/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/core/ui/alert/TextAlertDialog.kt
```diff
@@ -66,7 +63,7 @@ fun DestructiveTextAlertDialog(
         dismissButtonText = dismissButtonText,
         onPressPositiveButton = onPressPositiveButton,
         onDismissRequest = onDismissRequest,
-        confirmButtonFontColor = destructiveButtonFontColor
+        confirmButtonFontColor = destructiveButtonFont
```
FelberMartin, line 66: I'd suggest to use `ComponentColors.TextAlertDialog.destructiveButtonFont` here instead of directly importing the color on its own

FelberMartin (approved):
Looking good, I did not find any other hardcoded color values in the code anymore, good job!

core/ui/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/core/ui/alert/TextAlertDialog.kt
julian-wls: Yes, good catch!

PR #345, opened 2025-01-27
`Chore`: Fix window insets usage

FelberMartin:
### Problem Description
This is a followup to #320. This PR fixes the usage of WindowInsets in the context of the ReplyTextField.

### Changes
- Remove the consumption of `WindowInsets.systembars` and use the `navigationBarsPadding()` at the proper composables further down the UI tree.
- Remove scaffold in `MetisPostListHandler` as it messed with the padding values (probably because this scaffold was nested inside another scaffold)

### Steps for testing
Ensure that the paddings around the ReplyTextField are not behaving weirdly. Test with different navigation modes (gesture navigation, 3 buttonnavigation)


julian-wls (approved):
Did not observe any padding issues or weird behavior while testing.
Unseen-FAB in chat still looks placed correctly.

PR #354, opened 2025-01-28
`Improvement`: Allow users to navigate to exercise/lecture from corresponding channel

julian-wls:
### Problem Description

Currently, the web app allows users to navigate to the corresponding lecture/exercise from the channel. This is missing in Android.

### Changes

- A dropdown menu has been added that is shown when clicking on the info icon in the top bar.
- Navigating to the channel settings also works on title-click now.

### Steps for testing

1. Go to a exercise/lecture channel
2. Click the info icon and verify you can navigate to the corresponding exercise/lecture
3. Verify that there is no dropdown menu for dm chats, course-wide channels and exam channels


FelberMartin (changes requested):
Small but very handy change, I like it :D

I noticed that after clicking the "To Lecture/Exercise" button and then navigating back, the behavior is different between navigating with the Android back button and the TopAppBar back button. Tbh I am not sure why this is the case, maybe you can have a look. However I dont think this would be a dealbreaker if it cannot be fixed :)

julian-wls:
> I noticed that after clicking the "To Lecture/Exercise" button and then navigating back, the behavior is different between navigating with the Android back button and the TopAppBar back button. Tbh I am not sure why this is the case, maybe you can have a look. However I dont think this would be a dealbreaker if it cannot be fixed :)

I think we have the same issue when navigating from an exercise to a channel and trying to navigate back from the channel. I couldn't find a solution to this yet, but I'll open an issue to fix this in a follow up PR.

FelberMartin (approved):
lgtm now

eylulnc (approved):
Look good to me! 
Works as described and  I agree with the navigation problem.

PR #355, opened 2025-01-28
`Bugfix`: Fix crash when opening lecture attachments

julian-wls:
### Description

This PR fixes multiple problems related to lecture attachments sent into the chat:
- lecture attachments with the tag [attachment]...[/attachment] caused a crash when trying to open them using the pdf viewer
- lecture attachments [attachment], [lecture-unit] or [slide] that contained a space were not rendered correctly
- opening slides [slides] still caused an authorization error

### Changes

- Slides are now shown as images in the chat
- The lecture-unit regex has been changed to allow spaces
- the link has been ajusted for attachments [attachment]

### Steps for testing

1. Go to `Practical Course: Interactive Learning WS24/25` and open the `random` channel on TS1
2. See the last two messages by Artemis Test User 1 on Jan 28, 2025
3. Click on testset and verify the app is not crashing
4. Verify that the Interactive Learning pdf link is rendered correctly
5. Verify that the slide is rendered as an image 

### Discussion

With the recent change to maintain filenames when uploading pdfs in the chat there could be more occurrences of filenames that contain spaces. Should we adjust the regex for file uploads, too as part of this PR? What do you think @FelberMartin, @eylulnc?


FelberMartin (changes requested):
Tested and the described steps work as expected. Nicely done :D

Regarding the question of whether we should add this to the pdfs aswell, I think it would make sense.

I was wondering whether the space character is actually the only thing we need to encode/decode, so I created some weirdly named lecture attachements and posted them in the chat on TS1 > Test Course Martin Felber  in the one lecture chat. I found that some weirdly named attachements are not displayed correctly in the Android app. While these are super-edge cases, I think it still makes sense to maybe use the same regex as the webapp if possible.

image

julian-wls:
> I was wondering whether the space character is actually the only thing we need to encode/decode, so I created some weirdly named lecture attachements and posted them in the chat on TS1 > Test Course Martin Felber in the one lecture chat. I found that some weirdly named attachements are not displayed correctly in the Android app. While these are super-edge cases, I think it still makes sense to maybe use the same regex as the webapp if possible.

I just checked how this is done on the web app. It seems like the web app only uses regex to identify the raw pattern, but extracts the info such as title and url manually. (See [this file](https://github.com/ls1intum/Artemis/blob/c8c93d2ae27daca1b04ce626ad0a9605615d0664/src/main/webapp/app/shared/metis/posting-content/posting-content.components.ts#L238))

Moreover, I realized that the iOS app has the same problem and also doesn't support spaces.

I think in scope of this PR it would make sense to add the support for spaces in file-uploads and introduce a util class for the url en-/decoding. We could look more into the approach that is used on the web app and maybe replace our regex approach with this approach in the future in a separate PR. What do you think @FelberMartin?

SindiBuklaji (approved):
Tested on TS1 and it works as expected using Microsoft Edge 👍

FelberMartin:
> > I was wondering whether the space character is actually the only thing we need to encode/decode, so I created some weirdly named lecture attachements and posted them in the chat on TS1 > Test Course Martin Felber in the one lecture chat. I found that some weirdly named attachements are not displayed correctly in the Android app. While these are super-edge cases, I think it still makes sense to maybe use the same regex as the webapp if possible.
> 
> I just checked how this is done on the web app. It seems like the web app only uses regex to identify the raw pattern, but extracts the info such as title and url manually. (See [this file](https://github.com/ls1intum/Artemis/blob/c8c93d2ae27daca1b04ce626ad0a9605615d0664/src/main/webapp/app/shared/metis/posting-content/posting-content.components.ts#L238))
> 
> Moreover, I realized that the iOS app has the same problem and also doesn't support spaces.
> 
> I think in scope of this PR it would make sense to add the support for spaces in file-uploads and introduce a util class for the url en-/decoding. We could look more into the approach that is used on the web app and maybe replace our regex approach with this approach in the future in a separate PR. What do you think @FelberMartin?

Sounds reasonable to me 👍

eylulnc:
> > I was wondering whether the space character is actually the only thing we need to encode/decode, so I created some weirdly named lecture attachements and posted them in the chat on TS1 > Test Course Martin Felber in the one lecture chat. I found that some weirdly named attachements are not displayed correctly in the Android app. While these are super-edge cases, I think it still makes sense to maybe use the same regex as the webapp if possible.
> 
> I just checked how this is done on the web app. It seems like the web app only uses regex to identify the raw pattern, but extracts the info such as title and url manually. (See [this file](https://github.com/ls1intum/Artemis/blob/c8c93d2ae27daca1b04ce626ad0a9605615d0664/src/main/webapp/app/shared/metis/posting-content/posting-content.components.ts#L238))
> 
> Moreover, I realized that the iOS app has the same problem and also doesn't support spaces.
> 
> I think in scope of this PR it would make sense to add the support for spaces in file-uploads and introduce a util class for the url en-/decoding. We could look more into the approach that is used on the web app and maybe replace our regex approach with this approach in the future in a separate PR. What do you think @FelberMartin?

"Same here! Let's proceed with space support and the util class."

FelberMartin (approved):
lgtm

PR #362, opened 2025-01-29
`Improvement`: Unify search text fields

julian-wls:


eylulnc (commented):
This is nice improvement. I think one theme for search bar makes it elevated.

eylulnc:
I have 2 point I would like two discuss:

1 - Clear Icon 
The current background of the clear icon blends too much with the search bar, especially in certain themes (light and dark).

2 - Magnifier Icon and Cursor Placement
When the search bar is clicked, the cursor currently appears to the left of the magnifier icon, which can feel unintuitive.

Screenshot 2025-01-31 at 15 23 43

My suggestion would be r
1 - removing  the circle background entirely or switch to an inverted color scheme (e.g., a darker background with a lighter icon) to improve contrast and visibility. 

2 - Also for magnifier maybe adjustment the cursor position can be made after the magnifier icon for a more intuitive user experience. What do you guys think?

FelberMartin (approved):
I agree with Eylül's already mentioned points

Code lgtm

ahbitaqu (approved):
Works seamlessly here and looks great with the changes :)

eylulnc (approved):
Thanks for implementing the feedback points 😊. Its looking good and works as expected.

PR #367, opened 2025-01-30
`Bugfix`: Fix List Formatting

eylulnc:
### Problem Description
When using the markdown text field, revisiting an empty list item (e.g., a line with - or 1. ) causes the list prefix to be reinserted repeatedly with each keystroke. This disrupts the user experience by adding unwanted formatting while typing.

### Changes
 
Fix the bug

### Steps for testing

1. Navigate to a chat that supports markdown input.
2. Select one of the list options (either - for unordered or numbered list for ordered).
3. Write a few bullet points and leave the last list item empty.
4. Hit Enter and revisit the empty list item.
5. Write some text in the previously empty list item.
6. Verify that:
    -   The list prefix (-, 1., etc.) is not duplicated or corrupted.
    -   The new text is displayed correctly as part of the list.

### Screenshots
1 - Problem
2 - After Solution

Screenshot 2025-02-01 at 22 57 04
Screenshot 2025-02-01 at 23 00 05


FelberMartin (changes requested):
Nice catch, please also add a test for this bug. 

Please also revisit the "Steps for testing" and "Screenshots" sections in the PR description, at first glance I was super confused what should actually be tested here. Also adding a list formatting to the testing steps would improve the readability ;)

FelberMartin (changes requested):
Please add a unit test covering the bug scenario to `MarkdownListContinuationUtilTest`

julian-wls (approved):
Just tested it and it works as expected :+1: 
Code lgtm.

feature/metis/conversation/src/test/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/reply/MarkdownListContinuationUtilTest.kt
```diff
@@ -143,6 +143,109 @@ class MarkdownListContinuationUtilTest {
...
     }
 
+    @Test
+    fun `test WHEN revisiting an empty unordered list item THEN no repeated prefix is inserted`() {
+        // Initial text with an empty list item
+        val oldText = "- Test\n- \nNew text"
+
+        // Simulate the user placing the cursor in the empty list item and typing 'Text'
+        val newText = "- Test\n- Text\nNew text"
+        val cursor = newText.length  // Cursor should be after 'Text'
```
FelberMartin, line 153: The `cursor` values (also for the other tests) seem to be off. This will actually set the cursor at the very end of the text, after 'New text'

FelberMartin (changes requested):
Please verify whether the newly added tests actually cover the bug you want to fix with this PR. I ran the new tests with the pre-fix version of the `continueList()` implementation and all tests still pass. That means that the tests do not actually cover the bug. 

I recommend the following: 
1. Reset the state of `continueList` to the pre-fix implementation
2. Adjust the newly added to properly cover the bug scenario (this means some (or all) of the tests should fail)
3. Re-add the new, fixed `continueList` implementation
4. Verify that the tests are passing now :)

FelberMartin (approved):

PR #371, opened 2025-02-01
`Bugfix`: Multiple bug fixes related to receiving new post updates

### LINKED ISSUE #226: Unread counter increased while in in chat
### Describe the bug
When a user is in a chat and receives a new message and then navigates back to the conversation overview screen, the unread messages indicator is shown for this new message, even though the user has just read this message in the chat.

**To Reproduce**
1. Go to any chat with no unread messages
2. In the same chat, write  a message with another user from a different client (eg the web app)
3. In the android app, receive the new message
4. Navigate back and see in the ConversationOverview that the unread counter is increased for this chat

**Expected behavior**
The unread counter should not increase

FelberMartin:
### Problem Description
This PR fixes three bugs related to receiving new posts:
1. When receiving newly created messages in a chat the conversation's unread counter was not updated as expected. After reloading the unread counter would show these newly received messages as unread again. This PR closes #226 
2. Notifications about a chat were received when in that chat.
3. In the ChatList when receiving a new post, the ChatList did not scroll to the new post automatically

### Changes
 
- Added `markConversationAsRead` API
- Calling this API when the chat is visible
- Correct logic for checking whether notification target is visible
- Fix scrolling down when new post arrives in the ChatList

### Steps for testing
1. Activate notifications
1. Go to any chat with no unread messages
4. In the same chat, write  a message with another user from a different client (eg the web app)
5. In the android app, receive the new message 
  - ChatList automatically scrolls to new post
  - But no notification is received
5. Navigate back (and reload) and see in the ConversationOverview that the unread counter is still 0

### DIFF
--- a/feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/chatlist/MetisPostListHandler.kt
+++ b/feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/chatlist/MetisPostListHandler.kt
@@ -48,7 +48,6 @@ internal fun <T : Any> MetisPostListHandler(
 ) {
     val scope = rememberCoroutineScope()
 
-    var isScrolledDown by remember { mutableStateOf(false) }
     var prevBottomItem: T? by remember { mutableStateOf(null) }
 
     var hasNewUnseenPost by remember { mutableStateOf(false) }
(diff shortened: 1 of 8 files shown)


eylulnc (changes requested):
For me addressed problems still continues

I received a notification even though the message had already arrived.
The unread message count was inconsistent—sometimes it cleared, and other times it didn’t.
I also took a screen recording (sorry, it’s a bit long 😅).
Sometimes the message arrived very quickly, but other times it was delayed, or I had to manually trigger it. Additionally, the screen didn’t scroll to the new message automatically.

I'm not sure if the issue is on my side, related to the test server's performance, or due to the PR itself.

https://github.com/user-attachments/assets/111766e1-da14-4dfc-84ab-628ee557e99f

julian-wls:
> I received a notification even though the message had already arrived. The unread message count was inconsistent—sometimes it cleared, and other times it didn’t. I also took a screen recording (sorry, it’s a bit long 😅). Sometimes the message arrived very quickly, but other times it was delayed, or I had to manually trigger it. Additionally, the screen didn’t scroll to the new message automatically.

I just tested the changes on an emulator.
I experienced the following problems:

- The autoscroll was only working for the first message I sent
- The first message arrived within 2 seconds, but all following messages I sent were extremely delayed 
- I received some notifications after I closed the chat, but for the majority, I did not receive notifications.
- The unread counter seems to work fine for me most of the time

I realized that whenever a message was extremely delayed I received notifications although I was in the chat and the unread counter was not removed, however, the autoscroll also did not work for messages delivered quickly, it only worked once for me.

**Discussion**

In my opinion, most of the problems I experienced were not related to this PR. I'm not sure about the autoscroll, but under normal circumstances, the other features seem to work fine to me. How was your experience on a real device @FelberMartin?

FelberMartin:
Thanks to both of you for testing, scrolling to a new item did indeed not work as intended, this should be fixed now.

Regarding the other points, I am really not sure what is going on with the test servers atm, because I also experienced huge delays for larger courses. When I tested it with just a DM chat, all bugs seemed to be resolved for me. Maybe its just best for now to only test it in DM chats, as these huge delays really mess up everything. 
In case the huge delays (eg for notifcations) turn out to be not only a temporary issue, and we want to tackle not receiving notifications for chats the users is looking at, we would have to somehow record the timespans in which the user looks at a specific metisContext, and then when receiving the notification check whether at the post creation date, the related metisContext was visible and discard the notification. But this would be way out of scope imo (and hopefully not necassary)

@julian-wls @eylulnc Do you guys mind re-checking the fixes within a DM or small group chat? Note that notifications for unnamed groupchats do not work at the moment though, see #391

julian-wls (approved):
Awesome, everything worked for me now. I tested using a DM chat and didn't experience any issues or delays. I think we should merge this PR as every problem that previously occurred seemed to be server related. The changes of this PR did not cause any of the problems experienced earlier. With the scrolling working now, I think this PR is good to go :)

Code lgtm.

PR #375, opened 2025-02-02
`Feature`: Unread posts indicator

FelberMartin:
### Problem Description
When navigating to a chat with unread posts, the users had no clear indicator where the new unread posts begin. This PR adds an unread indicator inspired by slack and https://github.com/ls1intum/Artemis/pull/10018

### Changes
 
- Added unread indicator
- Added index to paged posts to correctly display the unread indicator

### Steps for testing
- Go to any chat with unread posts
- See that a visual indicator separates read from unread posts

### Screenshots
image

### Follow-up
**Known bug:**
  When the unread counter in the ConversationOverview increases due to a websocket update (the user is currently in the communication tab and a new post in a chat that is not open is sent), then the unread indicator is not displayed. This is due to the unread counter only being updated in the ConversationOverviewViewmodel. This bug will be tackled with #394


eylulnc (approved):
Code looks good.
Works as described, I did not encounter any problem other than the one mentioned in the description.

feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/chatlist/MetisChatList.kt
```diff
@@ -403,6 +411,39 @@ private fun NoPostsFoundInformation(
...
+        Text(
+            modifier = Modifier
+                .padding(end = 32.dp)
+                .align(Alignment.CenterEnd)
+                .background(
+                    color = color,
+                    shape = MaterialTheme.shapes.small
+                )
+                .padding(4.dp),
+            text = "New",
```
julian-wls, line 440: I think we should use a string resource here :)

FelberMartin:
I also experienced the second bug that you mentioned, I think this one is (as you mentioned) server related and also occurs on `develop`. This bug however could also be mitigated by storing the conversations locally. 

Overall I agree with you that the resulting behavior of these two bugs are quite disturbing, so I will merge this PR now, but include a feature toggle that deactives the feature for now. When the bugfixes are introduced, we then can also remove the feature toggle again.

julian-wls (approved):
I agree. Lgtm now.

PR #382, opened 2025-02-03
`Bugfix`: Fix PostItem scaling

eylulnc:
### Problem Description
When the user adjusts the system font size (for accessibility or personal preference), certain text elements—especially those rendered via our custom MarkdownText composable—do not scale properly. This leads to inconsistent UI: standard Compose Text elements scale up/down, but the markdown text remains the same size.

### Changes
 

1. Fixed Markdown Text Scaling

    Replaced TypedValue.COMPLEX_UNIT_DIP with TypedValue.COMPLEX_UNIT_SP in our MarkdownText composable.
    Ensured that any text size set on the underlying TextView is in scale‐independent pixels (sp), so it respects the user’s font‐size setting.

2. Removed Fixed DP Constraints

    Updated components (like HeadlineProfilePicture) that had hard‐coded dp sizes, allowing them to scale if desired or remain flexible in layout.

### Steps for testing
Go to the device’s Settings → Display → Font size and select “Largest.” ( Also among other font settings)
Return to App and open a communication channel
Check in the channel Post Header is no longer clipped and the post content size is also updated according to the font size you arranged.
Check that the profile picture also resizes to align with the new header height.

### Screenshots

Regular and Largest font size

Screenshot 2025-02-03 at 23 39 33
Screenshot 2025-02-03 at 23 39 00


eylulnc:
Emoji's should not be cutted out now. The icon button on the other hand is not someth
ing that can be scaled like the text in this context. There is of course a work around (scale components height or size by multiplying the fontScale) but I do not find it sustainable.

Screenshot 2025-02-04 at 19 46 56
Screenshot 2025-02-04 at 19 46 41

FelberMartin (approved):
Changes and code lgtm. Thank you for tackling this! :D

julian-wls (approved):
Really nice, works as expected!
Code lgtm

FelberMartin:
@eylulnc There is one merge conflict, once that one is fixed, feel free to add the "Ready to merge" label :)

PR #383, opened 2025-02-05
`Improvement`: Add elevation to top bar

julian-wls:
### Problem Description
As discussed after the last meeting the TopAppBar was not really separated from the content of the screen.

### Changes
This PR adds shadow elevation and a new background for the TopAppBar used in all course Scaffolds.
Multiple paddings had to be adjusted.

### Steps for testing
- Make sure all course related scaffolds still look good
### Screenshots

Screenshot_20250205_152512


FelberMartin (commented):
Nice, thanks for taking care of this!

Does anything speak against introducing a new class (eg `ArtemisTopAppBar`) that just provides the same function parameters as the default compose `TopAppBar`? In it, we could just wrap the TopAppBar in the elevated surface and apply our colors. We could then just use ArtemisTopAppBar everywhere in the app instead of TopAppBar. What do you think? @julian-wls

julian-wls:
> Does anything speak against introducing a new class (eg `ArtemisTopAppBar`) that just provides the same function parameters as the default compose `TopAppBar`? In it, we could just wrap the TopAppBar in the elevated surface, apply our colors, and pass down the remaining parameters. We could then just use ArtemisTopAppBar everywhere in the app instead of TopAppBar. What do you think? @julian-wls

Yes, I was thinking about this, too in the beginning and thought this would be overkill, but considering how big this PR turned out I totally agree. I'll make the changes soon.

feature/lecture-view/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/lectureview/LectureScreenUi.kt
```diff
@@ -179,7 +181,10 @@ internal fun LectureScreen(
                             overflow = TextOverflow.Ellipsis
                         )
                     },
-                    navigationIcon = { NavigationBackButton(onNavigateBack) }
+                    navigationIcon = { NavigationBackButton(onNavigateBack) },
+                    colors = TopAppBarDefaults.topAppBarColors(
```
FelberMartin, line 185: This should be redundant now :)

FelberMartin (approved):
Looking goood

PR #384, opened 2025-02-05
`Feature`: Add recent section to dashboard

julian-wls:

### DIFF
--- a/feature/dashboard/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/dashboard/ui/CourseOverviewViewModel.kt
+++ b/feature/dashboard/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/dashboard/ui/CourseOverviewViewModel.kt
@@ -11,6 +11,7 @@ import de.tum.informatics.www1.artemis.native_app.core.datastore.authToken
 import de.tum.informatics.www1.artemis.native_app.core.device.NetworkStatusProvider
 import de.tum.informatics.www1.artemis.native_app.core.model.Dashboard
 import de.tum.informatics.www1.artemis.native_app.feature.dashboard.service.DashboardService
+import de.tum.informatics.www1.artemis.native_app.feature.dashboard.service.DashboardStorageService
 import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.SharingStarted
 import kotlinx.coroutines.flow.StateFlow
@@ -25,6 +26,7 @@ import kotlin.coroutines.EmptyCoroutineContext
  */
 internal class CourseOverviewViewModel(
     private val dashboardService: DashboardService,
+    private val dashboardStorageService: DashboardStorageService,
     accountService: AccountService,
     serverConfigurationService: ServerConfigurationService,
     networkStatusProvider: NetworkStatusProvider,
--- a/feature/dashboard/src/debug/kotlin/de/tum/informatics/www1/artemis/native_app/feature/dashboard/DashboardScreenshots.kt
+++ b/feature/dashboard/src/debug/kotlin/de/tum/informatics/www1/artemis/native_app/feature/dashboard/DashboardScreenshots.kt
@@ -26,8 +26,12 @@ import de.tum.informatics.www1.artemis.native_app.core.ui.ScreenshotFrame
 import de.tum.informatics.www1.artemis.native_app.core.ui.remote_images.LocalArtemisImageProvider
 import de.tum.informatics.www1.artemis.native_app.core.ui.test.ArtemisImageProviderStub
 import de.tum.informatics.www1.artemis.native_app.feature.dashboard.service.DashboardService
+import de.tum.informatics.www1.artemis.native_app.feature.dashboard.service.DashboardStorageService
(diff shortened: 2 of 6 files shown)


feature/dashboard/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/dashboard/ui/CoursesOverview.kt
```diff
@@ -167,6 +179,43 @@ internal fun CoursesOverview(
...
             )
 
+            Row(
+                modifier = Modifier
+                    .fillMaxWidth()
+                    .padding(16.dp),
+                horizontalArrangement = Arrangement.spacedBy(8.dp),
+                verticalAlignment = Alignment.CenterVertically
+            ) {
+                Box(modifier = Modifier.weight(1f)) {
```
FelberMartin, line 189: This Box should not be required and instead using the `weight(1f)` modifier on the `BasicSearchText` should be enough. I checked and found that the Box actually _is_ required because the modifier passed to `BasicSearchTextField` does not get applied to its root element.  
I think this should be fixed, either in this PR if it doesn't mess up the other usages of `BasicSearchTextField` too much, or in a follow-up.

feature/dashboard/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/dashboard/ui/CoursesOverview.kt
```diff
@@ -167,6 +179,43 @@ internal fun CoursesOverview(
                 surveyHintService = surveyHintService
             )
 
+            Row(
```
FelberMartin, line 182: Create a new composable for this (something like `SearchAndOrderRow`)

FelberMartin (changes requested):
Awesome, I like this change!

I have a few code comments and one general remark:
When using the sorting option or the search, it looks like the LazyColumn tries to remember which element the user has been looking at previously and adjusts the scroll state to still show that item after reordering.


https://github.com/user-attachments/assets/42d236b7-db20-4611-8495-c710fa6c53ff
Here the "General Testing Course" is the topmost course and in focus, and when I press the change order button, I would expect to see the (previous) bottommost element ("Test Course Matthias Linhuber"). But instead I still see the "General Testing Course" and have to scroll to the top of the list to see the "Test Course Matthias Linhuber"

feature/dashboard/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/dashboard/ui/CourseOverviewViewModel.kt
```diff
@@ -36,33 +40,146 @@ internal class CourseOverviewViewModel(
...
                 }
+            }.collect {
+                _dashboardState.value = it
             }
         }
-            //Store the loaded dashboard, so it is not loaded again when somebody collects this flow.
-            .stateIn(viewModelScope + coroutineContext, SharingStarted.Eagerly, DataState.Loading())
+    }
+
+    fun reorderCourses() {
```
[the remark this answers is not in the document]
julian-wls: Yes, you're right I used LaunchedEffect for testing purposes and forgot to add the flow that observes course accesses.

julian-wls:
> I have a few code comments and one general remark: When using the sorting option or the search, it looks like the LazyColumn tries to remember which element the user has been looking at previously and adjusts the scroll state to still show that item after reordering.
> 
> https://github.com/user-attachments/assets/42d236b7-db20-4611-8495-c710fa6c53ff Here the "General Testing Course" is the topmost course and in focus, and when I press the change order button, I would expect to see the (previous) bottommost element ("Test Course Matthias Linhuber"). But instead I still see the "General Testing Course" and have to scroll to the top of the list to see the "Test Course Matthias Linhuber"

I agree, this is really unintuitive and feels strange, I added function to always save the last list position before changing the sorting and doing filtering. This way after doing a reordering the list can revert to its previous position. It should all work as expected now. Could you check again, @FelberMartin?

feature/dashboard/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/dashboard/ui/CourseList.kt
```diff
@@ -38,16 +49,53 @@ fun CourseList(
...
         modifier = modifier.consumeWindowInsets(WindowInsets.navigationBars),
         columns = GridCells.Fixed(columnCount),
+        state = courseListState,
         verticalArrangement = Arrangement.spacedBy(Spacings.CourseItem.gridSpacing),
         horizontalArrangement = Arrangement.spacedBy(Spacings.CourseItem.gridSpacing),
         contentPadding = Spacings.calculateEndOfPagePaddingValues()
     ) {
+        if (recentCourses.isNotEmpty()) {
+            item(span = { GridItemSpan(columnCount) }) {
+                SectionHeader(stringResource(R.string.course_overview_recently_accessed_section))
```
FelberMartin, line 59: Very minor, but the seciton headers would also benefit from `Modifier.animateItem()` :D

FelberMartin (approved):
Looking very nice now!

PR #405, opened 2025-02-11
`Chore`:  Introduce  ArtemisContext and ArtemisContextBasedService

FelberMartin:
### Problem Description
Currently we have a lot of code duplication in the service implementations. Most prominently, the serverUrl and authToken is passed down to (almost) every single API call. This clutters the code, especially in the ViewModels.

This PR is the first of several planned PRs for introducing the new concept of an `ArtemisContext` and an `ArtemisContextBasedService` that does not need a serverUrl or authToken as function parameters. To keep the PR size in a manageable scale, there will be follow-up PRs dealing with the introduction of this `ArtemisContext` to other services as well. For this first PR, I only applied the necassary changes to the Services defined in the `core:data` module.

### Changes
 
- Introduced 
  - `ArtemisContext`: Simple data class to keep track of serverUrl and authToken
  - `ArtemisContextProvider`: Provides a flow for tracking the latest context
  - `ArtemisContextBasedService`: Interface used by service definitions to indicate that the serverUrl and authToken are handled internally and not needed as function parameters
  - `ArtemisContextBasedServiceImpl`: Implementation of the interface above. Provides convenient methods for API requests (taking care of the serverUrl and authToken in the background)

### Steps for testing
See the affected services in the commit messages and confirm that related functionality in the app still works


julian-wls (changes requested):
Really nice changes this makes already makes it a lot cleaner. I tried to go through all of the screens of the app and noticed that the pull to refresh is not working anymore in the Exercise Screen, it does not happen on `develop`. Could you take a look?

Couldn't find issues in your code.

FelberMartin:
Nice catch, this should be fixed now :)

julian-wls (approved):
Didn't notice any issues. Everything seems to work now.

PR #408, opened 2025-02-12
`Bugfix`: Fix emoji icon scaling

julian-wls:
### Problem Description

Currently, the icon for the add emoji button seems a bit too big and not aligned, see screenshot:
image

It looks fine when the text size is bigger than usual, but not with the normal scaling.

### Changes

This PR changes the height of the emoji to use .sp to be affected from the font changes and therefore be aligned with the other emoji chips.

### Steps for testing

Make sure it looks good before and after changing the font size.


feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/post/PostItem.kt
```diff
@@ -592,14 +591,17 @@ private fun StandalonePostFooter(
                     modifier = modifier
                         .background(color = PostColors.EmojiChipColors.background, CircleShape)
                         .clip(CircleShape)
+                        .then(Modifier.align(Alignment.CenterVertically))
```
FelberMartin, line 594: I am not sure, is there any difference to using `.align(Alignment.CenterVertically)` directly?

FelberMartin (changes requested):
Nice change, there are two issues I noticed:
1. For small font sizes, the emojis and counter seem to be not centerred anymore
image
2. For very large font sizes, the icon is still a little bit smaller than the emojis
image

feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/post/PostItem.kt
julian-wls: .then was needed to wait for the icon to adjust it's true size since we are using .sp now.

julian-wls:
> Nice change, there are two issues I noticed:
> 
> 1. For small font sizes, the emojis and counter seem to be not centered anymore

> 2. For very large font sizes, the icon is still a little bit smaller than the emojis

Good catch, I managed to fix it for the very small font sizes.
For the large font sizes, I'm not quite sure, when selecting the largest it seems aligned correctly for me:
image

FelberMartin:
> For the large font sizes, I'm not quite sure, when selecting the largest it seems aligned correctly for me

Hm interesting, I re-tested on both foldable and physical device, and in both cases the icon size was off for the largest font size🤔 

https://github.com/user-attachments/assets/23f79c5b-8fe9-4400-a09b-a6ceea92edd5

FelberMartin:
Uh good to know. I just checked in the Android settings, there are two different scales. Display size apparently controls the `.dp` scalings, and font size the `.sp` scaled elements.
image

julian-wls:
Oh, I only had 4 steps on the font scale (API 33):

image

I tested on API 35 now, and it works with the ultra big font size now :) 
The solution was to change the padding of the box to also use .sp.

feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/post/PostItem.kt
julian-wls: Yes, true. I just checked and we can get rid of `align(Alignment.CenterVertically)
`

FelberMartin (approved):

PR #418, opened 2025-02-14
`Feature`: Add FAQ categories

### LINKED ISSUE #399: FAQ: Add categories
Faqs can have categories assigned to them. Users should be able to see these categories and filter by them

FelberMartin:
### Problem Description
Categories can be added to FAQs to easily group them and allow users to quickly find what they are looking for. 

This PR closes #399 

### Changes
 
- Display FAQ categories in both the Overview and Details screen
- Allow users to filter FAQs by category
- Added UI tests

### Steps for testing
Go to a course with FAQs (eg TS2 > Test Course Martin Felber)
- Check that the FAQ categories are displayed in both the Overview and details screen
- Check that filtering by category works as expected

### Screenshots
image


eylulnc (changes requested):
I think its looking really nice and works a sexpected.

One suggestion on my side is 

Screenshot 2025-02-15 at 16 41 03

This is looking cut out a lot.
Maybe the categories and read more should not display in the same line.

feature/faq/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/faq/ui/shared/FaqCategoryChip.kt
```diff
@@ -0,0 +1,120 @@
...
+        is FaqCategoryChipConfig.Filter -> testTagForCategoryFilterChip(faqCategory)
+    }
+
+    FilterChip(
+        modifier = modifier
+            .testTag(testTag),
+        enabled = filterConfig != null,
+        selected = filterConfig?.isSelected ?: false,
+        onClick = filterConfig?.onClick ?: {},
+        colors = FilterChipDefaults.filterChipColors().copy(
```
julian-wls, line 94: I'm not 100%, but I think we don't need the `.copy` here. I think it's possible to directly modify `filterChipColors(disabledContainerColor = ...)`

julian-wls (commented):
Really nice changes :)
 
I agree with @eylulnc.
I think instead of using a row we could either use a flow row (I would only use a flow row if it is possible to set the max amount of rows to 2 though) or move the button below the category row.

FelberMartin:
I addressed the code comment and added the categories as a FlowRow below the title as discussed on Tuesday.
(Btw, I did not specify a `maxLines = 2` for the FlowRow, because the maximum number of categories per faq is 2)

eylulnc (approved):
I really like this approach! It's much smoother when clicking on the FAQ, and having the filters in the same place.

julian-wls (approved):
Just tested it and works as expected.
I agree, if there are never more than two chips we don't need the `maxLines = 2`

Looks really nice now, especially the animations :)
Code lgtm!

feature/faq/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/faq/ui/overview/FaqOverviewUi.kt
```diff
@@ -138,6 +159,16 @@ private fun FaqOverviewBody(
     Column(
         modifier = modifier,
     ) {
+        if (filterChips.size > 1) {
+            ConfiguredFaqCategoryChipRow(
+                modifier = Modifier
+                    .fillMaxWidth()
+                    .padding(bottom = 8.dp)
```
julian-wls, line 166: Maybe one more comment:

I think we should increase the space between the top bar and the category row. Maybe 8.dp or even 16.dp. What do you think @FelberMartin?

FelberMartin: Good catch, I did not recheck the UI after merging the SearchBar changes :D

PR #419, opened 2025-02-14
`Chore`: Use ArtemisContext batch2

FelberMartin:
### Problem Description
This is a follow-up to #405 

### Changes
 
This PR covers all files in the following modules:
- core
- feature:
  - course-view
  - dashboard
  - exercise-view

Additionally I introduce a new convenience method `performAutoReloadingNetworkCall`

### Steps for testing
Ensure the changed services still work without problems (see commit messages for affected services)


feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/ConversationViewModel.kt
```diff
@@ -222,20 +216,23 @@ internal open class ConversationViewModel(
...
-    ) { serverUrl, authToken, course, _ ->
-        retryOnInternet(networkStatusProvider.currentNetworkStatus) {
-            accountDataService.getAccountData(
-                serverUrl = serverUrl,
-                bearerToken = authToken
-            )
-                .bind { it.isAtLeastTutorInCourse(course = course.orThrow()) }
+    ) { accountDataState, courseDataState ->
+        accountDataState.bind {
+            it.isAtLeastTutorInCourse(course = courseDataState.orThrow())
```
julian-wls, line 233: Maybe we should not use `courseDataState.orThrow()` here. This might be the problem.

julian-wls (changes requested):
Nice change

I noticed two issues:

- When clicking on the channel link in any exercise the corresponding conversation is not opened anymore and instead the view reloads.
- When trying to open conversations the app crashes (This did not happen consistently, but rather randomly): 
image

See below:

FelberMartin:
After fixing the merge conflicts and removing the `orThrow()` I could no longer reproduce "When clicking on the channel link in any exercise the corresponding conversation is not opened anymore and instead the view reloads." can you confirm @julian-wls ?

julian-wls (commented):
There seems to be an error in LecutureScreenUi, Line 229 in LinkBottomSheet. I think you removed serverUrl and authToken and added artemisContext. The LinkBottomSheet still seems to use the old parameters in LectureScreenUi. This was probably caused by merging the lecture screen PR. Could you take a look?

julian-wls (approved):
Tested it and this time I didn't experience any issues. 
Code lgtm

PR #421, opened 2025-02-15
`Feature`:  Force Update

eylulnc:
### Problem Description
There is currently no way to handle force-updates in our Android app in a streamlined way. If the user is on an outdated version, they continue using the app without being prompted to update. We also had some inconsistencies in how the app checks for new versions from the server.

### Changes
 

- Introduced a new UpdateRepository that retrieves the latest server version and compares it against our local versionCode.
- Modified MainActivity to collect a Flow from the UpdateRepository. Whenever an update is required, we navigate the user to a new “UpdateScreen.”
- If the server request fails. If the request cannot be completed, we show no update requirement.

### Steps for testing

**1. Initial Setup**
- Install the app from scratch.
- Set up a local server that reports the latest version code.
Note: To test with [PR #10339](https://github.com/ls1intum/Artemis/pull/10339), ensure the server correctly provides the app version.

**2. Test Cases**
_Case 1: No Update Required_
The update screen should NOT be displayed if the server version matches the local app version.

Steps:
1. Ensure the server reports a version equal to the current app version.
2. Launch the app.
3. Verify that no update prompt appears.

_Case 2: Update Required_
The update screen should be displayed if the server requires a newer version, and clicking the "Download" button should navigate to the Google Play Store.

Steps:
1. Locally modify the server response to return a higher version than the current app version. ( Please test it for both 1.x.x and 1.x.x-prod)
2. Launch the app.
3. Verify that the UpdateScreen appears.
4. Click the "Download" button.
5. Confirm that the button redirects to the Google Play Store.

If the app cannot reach the server, it should not force an update.

### Screenshots

Screenshot 2025-02-15 at 16 01 27
Screenshot 2025-02-18 at 17 34 39


feature/force-update/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/force_update/repository/UpdateRepository.kt
```diff
@@ -0,0 +1,101 @@
...
+            }
+
+            //Do a new call to /management/info
+            val token = accountService.authToken.first()
+            val response = updateService.getLatestVersion(latestServerUrl, token)
+
+            when (response) {
+                is NetworkResponse.Response -> {
+                    // parse min version from “compatible-versions”
+                    val minVersion: Int = response.data ?: 0
```
eylulnc, line 63: If response is null or etc. Should we check if in the local is there any version saved before doing this?

feature/force-update/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/force_update/ui/UpdateScreen.kt
```diff
@@ -0,0 +1,100 @@
...
+                painter = painterResource(R.drawable.ic_launcher_foreground),
+                contentDescription = "Update Icon",
+                modifier = Modifier
+                    .size(UpdateScreenConstants.ImageSize)
+                    .align(Alignment.CenterHorizontally),
+            )
+
+            Text(
+                text = stringResource(id = R.string.update_screen_new_update_available),
+                style = MaterialTheme.typography.titleLarge.copy(
```
julian-wls, line 60: For some reason the color of the two following text is not changing in dark mode:
image

Could you add `color = MaterialTheme.colorScheme.onBackground` for both?

feature/force-update/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/force_update/UpdateScreenConstants.kt
```diff
@@ -0,0 +1,11 @@
+package de.tum.informatics.www1.artemis.native_app.feature.force_update
+
+import androidx.compose.ui.unit.dp
+
+object UpdateScreenConstants {
```
julian-wls, line 5: I think it would be good to stay consistent and move this object to `Spacings` (see ui module, Spacings.kt)

feature/force-update/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/force_update/ui/UpdateScreen.kt
```diff
@@ -0,0 +1,100 @@
...
+    onDownloadClick: () -> Unit,
+    currentVersion: String,
+    minVersion: String
+) {
+    Box(
+        modifier = Modifier
+            .fillMaxSize()
+            .background(color = MaterialTheme.colorScheme.background)
+            .imePadding()
+            .padding(bottom = UpdateScreenConstants.SpacingMedium),
```
julian-wls, line 40: Then we could use it as `Spacings.UpdateScreen.medium` of something like that and apply it to the other paddings, too :)

feature/force-update/src/main/res/values/update_screen_strings.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="update_screen_new_update_available">New update available</string>
+    <string name="update_screen_download_message">Please update Artemis to the latest version to ensure you have a smooth experience. You can download the latest version from the Play Store.</string>
+    <string name="update_screen_downlaod_button">Download</string>
+    <string name="update_screen_version_info">
+    You are currently using version %1$s while the oldest version currently supported is %2$s.
```
julian-wls, line 7: I think instead of using currently twice, let's just say `You are using version %1$s, while the oldest version currently supported is %2$s.`

julian-wls (changes requested):
Nice changes, I just tested both test cases.

Test Case 1:
I started the app after switching to this branch and everything worked as expected, no update screen was shown.

Test Case 2:
The update screen was shown and the redirect button to the play store works as expected.

Code looks generally good to me,
just some suggestion for improvement, see below:

feature/force-update/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/force_update/ui/UpdateScreen.kt
```diff
@@ -0,0 +1,102 @@
...
+                color = MaterialTheme.colorScheme.secondary,
+                textAlign = TextAlign.Center,
+            )
+
+            Spacer(modifier = Modifier.weight(0.1f))
+
+            Button(
+                onClick = onDownloadClick,
+                modifier = Modifier.fillMaxWidth(),
+                shape = RoundedCornerShape(Spacings.UpdateScreen.buttonRadius)
```
FelberMartin, line 94: I assume the custom shape here is to align with the iOS design? I'd argue to stick with the Material3 standard button shape, as all other buttons in the Android app currently use that as well

feature/force-update/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/force_update/ui/UpdateScreen.kt
```diff
@@ -0,0 +1,102 @@
...
+                    R.string.update_screen_version_info,
+                    currentVersion,
+                    minVersion
+                ),
+                style = MaterialTheme.typography.bodyMedium,
+                color = MaterialTheme.colorScheme.secondary,
+                textAlign = TextAlign.Center,
+            )
+
+            Spacer(modifier = Modifier.weight(0.1f))
```
FelberMartin, line 89: It does not make a technical difference, but simply using `weight(1f)` should have the same effect here. Or is there any specific reasoning behind using `0.1f`?

feature/force-update/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/force_update/ui/UpdateScreen.kt
```diff
@@ -0,0 +1,102 @@
...
+    onDownloadClick: () -> Unit,
+    currentVersion: String,
+    minVersion: String
+) {
+    Box(
+        modifier = Modifier
+            .fillMaxSize()
+            .background(color = MaterialTheme.colorScheme.background)
+            .imePadding()
+            .padding(bottom = Spacings.UpdateScreen.medium),
```
FelberMartin, line 40: Instead of `.imePadding().padding(bottom = Spacings.UpdateScreen.medium),` we probably want to use `systemBarsPadding()`. There is no way to pop up the software keyboard on this screen, right?

feature/force-update/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/force_update/ui/UpdateScreen.kt
```diff
@@ -0,0 +1,102 @@
...
+import de.tum.informatics.www1.artemis.native_app.core.ui.Spacings
+import de.tum.informatics.www1.artemis.native_app.feature.forceupdate.R
+
+@Composable
+fun UpdateScreen(
+    onDownloadClick: () -> Unit,
+    currentVersion: String,
+    minVersion: String
+) {
+    Box(
```
FelberMartin, line 35: Instead of this box, we should use a Scaffold here imo. This will also eliminate the need to set the background color and text colors by hand

feature/force-update/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/force_update/ui/UpdateNavGraph.kt
```diff
@@ -0,0 +1,33 @@
...
+
+import androidx.navigation.NavController
+import androidx.navigation.NavGraphBuilder
+import androidx.navigation.compose.composable
+
+private const val UPDATE_SCREEN_ROUTE = "update_screen/{currentVersion}/{minVersion}"
+
+fun NavController.navigateToUpdateScreen(currentVersion: String, minVersion: String) {
+    navigate("update_screen/$currentVersion/$minVersion") {
+        popUpTo(graph.startDestinationId) { inclusive = false }
```
FelberMartin, line 11: To align with the rest of the project, we should use serialization based navigation here as well. You can see how a screen route with arguments is defined eg in `CourseUiScreen`

core/model/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/core/model/server_config/ProfileInfo.kt
```diff
@@ -31,4 +31,16 @@ data class ProfileInfo(
...
+}
+
+@Serializable
+data class CompatibleVersions(
+    val android: AndroidCompatibleVersions? = null
+)
+
+@Serializable
+data class AndroidCompatibleVersions(
+    val min: String? = null,
```
FelberMartin, line 44: Being a little bit more explicit here and naming the value `minRequired` would make the logic more readable. You can use `@SerialName("min") to still serialize correctly

feature/force-update/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/force_update/service/UpdateService.kt
```diff
@@ -0,0 +1,46 @@
...
+import de.tum.informatics.www1.artemis.native_app.core.model.server_config.ProfileInfo
+import io.ktor.client.call.body
+import io.ktor.client.request.get
+import io.ktor.http.ContentType
+import io.ktor.http.appendPathSegments
+import io.ktor.http.contentType
+
+interface UpdateService {
+    suspend fun getLatestVersion(
+        serverUrl: String,
```
FelberMartin, line 16: Can't we use the `ServerProfileInfoService` instead? They both call the same API endpoint. We could move `ServerProfileInfoService` out of the login module into the core

FelberMartin (changes requested):
Nice, looking forward to getting this merged as soon as possible :D

Besides the code comments, one general remark: I think this feature is like the prime example where we would _really_  benefit from writing tests. Imo there should be tests for the `isVersionGreater` function and also for the update logic itself (the logic returning an UpdateResult). 

In case adding these tests would delay the feature for too long, I am also fine with merging the feature for now just with manual testing, though

feature/force-update/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/force_update/ui/UpdateScreen.kt
eylulnc: Nope, there shouldn't be a case.

feature/force-update/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/force_update/ui/UpdateScreen.kt
eylulnc: Generally 0.1f takes less space but in this context I dont think it makes any diffrence so I updated to 1f

eylulnc:
> Nice, looking forward to getting this merged as soon as possible :D
> 
> Besides the code comments, one general remark: I think this feature is like the prime example where we would _really_ benefit from writing tests. Imo there should be tests for the `isVersionGreater` function and also for the update logic itself (the logic returning an UpdateResult).
> 
> In case adding these tests would delay the feature for too long, I am also fine with merging the feature for now just with manual testing, though

I add test for isVersionGreater but the other will require more time

feature/force-update/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/force_update/service/UpdateService.kt
```diff
@@ -0,0 +1,32 @@
+package de.tum.informatics.www1.artemis.native_app.feature.force_update.service
+
+import de.tum.informatics.www1.artemis.native_app.core.data.NetworkResponse
+import de.tum.informatics.www1.artemis.native_app.core.datastore.ServerProfileInfoService
+
+interface UpdateService {
+    suspend fun getLatestVersion(
+        serverUrl: String,
+        authToken: String
```
FelberMartin, line 9: We do not need the authToken anymore

FelberMartin (changes requested):
The tests are fine from my side for now, please create a follow-up issue so we do not forget about this.

One remark for writing the remaining tests: We could create a new `object UpdateUtil` with functions like `isVersionGreater`, `isTimeToCheckUpdate(now: Long)` and `createUpdateResultBasedOnServiceResponse(response: NetworkResponse<String?>)` and then we could more easily and independently test the logic of these functions. This way, we could avoid mocking all the parameters of `UpdateRepository` :D

feature/force-update/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/force_update/repository/UpdateRepository.kt
```diff
@@ -0,0 +1,144 @@
...
+        return false
+    }
+
+    private val currentVersionNormalized = normalizeVersion(version)
+
+    /**
+     * Checks for an update whenever the server URL changes or every 2 days.
+     * Instead of using an integer version code, we compare version strings.
+     */
+    fun checkForUpdate(): Flow<UpdateResult> =
```
FelberMartin, line 62: This could be a `val updateResultFlow`

feature/force-update/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/force_update/repository/UpdateRepository.kt
```diff
@@ -0,0 +1,144 @@
...
+                    forceUpdate = true,
+                    currentVersion = currentVersionNormalized,
+                    minVersion = storedServerVersion
+                )
+            }
+
+            // If it's not yet time to re-check (less than 2 days since last check), then assume no update.
+            if (!isTimeToCheckUpdate()) {
+                return@mapLatest UpdateResult(
+                    updateAvailable = false,
```
FelberMartin, line 83: We could create a `val noUpdateNeeded` for this UpdateResult and then also return it in case of network failures

eylulnc:
> The tests are fine from my side for now, please create a follow-up issue so we do not forget about this.
> 
> One remark for writing the remaining tests: We could create a new `object UpdateUtil` with functions like `isVersionGreater`, `isTimeToCheckUpdate(now: Long)` and `createUpdateResultBasedOnServiceResponse(response: NetworkResponse<String?>)` and then we could more easily and independently test the logic of these functions. This way, we could avoid mocking all the parameters of `UpdateRepository` :D

I added a Util class and  also test cases for the functions as well.

I also manually tested out but **please** check the test cases again just to be sure.
If we want to add more test cases I can create a follow up.

eylulnc:
@FelberMartin @julian-wls I also noticed that when a network request fails, we fall back to '0.0.0'. Should we first check if we have anything stored? If so, do we want to address this now or in a follow-up?

FelberMartin (approved):
Nice, code looks good to me now :D

Testing both scenarios worked as described

PR #423, opened 2025-02-19
`Feature`: Channel Filter

julian-wls:
### Problem Description

Currently, the channel filter, which is available on the iOS app, is missing on Android.

### Changes

This PR adds the channel filter to the Android app using the material3 FilterChips.
- A new sealed class ConversationFilter has been added to ConversationOverviewUtils, which provides objects for all four possible filters.
- The ConversationOverviewModel has been adjusted to check which filters should be shown and holds the currently selected filter.
- The ConversationCollections include the filter logic
- The recent section has been replaced and recent channels are only shown when the recent filter has been selected.
- Improvements and animations have been added the ConversationOverviewBody

### Steps for testing

1. Log in as tutor
2. Go to any course and verify that the unresolved filter is shown.
3. Go to a course that has currently ongoing exercises and unread messages (Test Course Julian Waluschyk on TS1) and verify that all filter options are shown.
4. Test the filter functionality and check that the filter is still selected after navigating to a conversation and going back to the overview (except for unread message after the last unread conversation was clicked)

### Screenshots

channel_filter_small


eylulnc (changes requested):
Look really nice 😍 I loved it and works as expected.

I have some comment related to paddings and colors:

1 - Padding is slightly of between top and bottom of the filters.
Screenshot 2025-02-20 at 11 23 13

2 - I think in the colors are too vibrant compare to iOS, also 
Screenshot 2025-02-20 at 11 28 14
for me purple is not that balanced in dark mode but it could also be a problem related to me 😂

Also related to the saved message do we have any reason to  keep them in every filter?

FelberMartin (approved):
Awesome change, works as expected and code also lgtm!

julian-wls:
I just made the following changes:

> 1 - Padding is slightly of between top and bottom of the filters.
I tried to change the padding to be equal for top and bottom.

> 
> 2 - I think in the colors are too vibrant compare to iOS, also for me purple is not that balanced in dark mode but it could also be a problem related to me 😂
 I changed purple now to be a little bit brighter.

> Also related to the saved message do we have any reason to keep them in every filter?
I'm not sure I think this was intended by @FelberMartin. What's you opinion on this @FelberMartin 

@eylulnc could you check again?

eylulnc (approved):
Thanks for the adaptations. Looks good.

PR #444, opened 2025-02-28
`Feature`: Add version code to push notification registrattion

FelberMartin:
### Problem Description
With a recent API change, the server expects the app version to be send along the push notification registration call. See https://github.com/ls1intum/Artemis/pull/10384

### Changes
 
- Introduce `AppVersion` and `AppVersionProvider` to easily access the app version in the app without having to pass it all the way from the `app` module
- Add `NormalizedAppVersion` to be used in modified API call and in the force-update feature
- Added unit tests

### Steps for testing
See that updating the push notification settings still works as expected


eylulnc (approved):
Code lgtm.
 Notification settings works as before.

julian-wls (approved):
Works as expected.
Code lgtm

Issue #379, opened 2025-02-03
Investigate back navigation after opening exercise/lecture link from channel


julian-wls:
### Describe the bug

As mentioned in [this conversation](https://github.com/ls1intum/artemis-android/pull/354#pullrequestreview-2579138817) of #354 after navigating from a channel to an exercise or a lecture the back button does not work as expected. The behavior is not the same for the Android back button and the top bar back button.

**To Reproduce**
1. Go to a channel that has a corresponding exercise/lecture
2. Click on the info button
3. Click on Go to exercise/lecture
4. Try to navigate back using both the Android back button and the top bar back button
5. See that the behavior is different

### Additional context
This issue was introduced with #354.

PR #447, opened 2025-02-28
`Feature`: Mention FAQ in chat

### LINKED ISSUE #401: FAQ: Support mention in chat
Users should be able to mention faqs in the chat. When clicking on a faq mention in the chat, the user should be taken to the corresponding faq details page.

FelberMartin:
### Problem Description
FAQs can be mention in a post, just as eg lectures or exercises. When clicking on such a mention link, the user gets redirected to the faq's details page.

This PR closes #401 

### Changes
 
- Added deeplinks markdown transformation for mentioned faqs
- Add faqs to autoCompletion options
- Refactorings

### Steps for testing
1. Go to a course with FAQ enabled (eg TS2 > Test Course Martin Felber)
2. Mention a faq in a post by using the autocompletion dialog
3. Click on the faq link in the new post
4. Get redirected to the faqs details page

### Screenshots

https://github.com/user-attachments/assets/c57f1a0e-60ee-4d17-867e-c6fab4224270


eylulnc (changes requested):
Code lgtm.
Also works as described too. Nice work 😊

I have one slight suggestion but I dont have a strong opinion about it.

I think the padding between the question Mark icon and FAQ is slightly too much it looks unbalanced. 
Screenshot 2025-03-01 at 15 38 34

Also one thing related to navigation  the back button inside of the app navigate me to course list instead of the chat.

julian-wls (approved):
Really nice change, I could only test it with one FAQ, but it all worked as expected.
Regarding the Question mark icon I agree with @eylulnc, but I know how hard it is to work with these icons in markdown, so I think if there is no easy way to fix this it should be fine, too.

Code lgtm! I like the AutoCompletionUseCase, makes it a lot cleaner :)

FelberMartin:
> Also one thing related to navigation the back button inside of the app navigate me to course list instead of the chat.

This is related to #379 

Regarding the icon padding issue, as Julian pointed out, this is quite hard to do, we would have to modify the svg path in order to make that work :/

eylulnc (approved):
Thanks for the elaboration 😊

PR #448, opened 2025-02-28
`Feature`: Show forwarded messages

julian-wls:
### Problem Description

Forwarded messages are currently not rendered by the app, and that's why there are sometimes posts with no content in the chat.

### Changes

This PR introduces the rendering of forwarded messages by making the required requests and adding logic to correctly match the forwarded posts to their destination posts. 
The following changes have been made:

- extracting and loading functionality for forwarded messages in the PageSource of MetisChatList
- Replaced PostItemViewType with already existing ChatListItem, and expanding ChatListItem to feature more sub-classes and store forwarded messages
- Added ForwardedMessageColumn with rendering functionality and conversation matching
- Changed PostItemMainContent to not show role badges anymore as author role was always null (also for saved messages)
- added visibility tests for forwarded messages
- fixed bugs

Note: 
- This PR does **not** introduce the functionality to create forwarded messages, as this will be done in a follow up PR to not make this PR bigger as it already is, see #460 
- Forwarded messages are currently not rendered in saved posts, this should also be done in a follow up PR, see #459 

### Steps for testing

1. Go to a chat that includes forwarded messages or create some in the web app
2. Make sure the forwarded messages are shown correctly
3. Check if the the source conversation is shown correctly and clickable if public
4. Change the content of a forwarded message and make sure the changes are reflected by the app
5. Delete the source post of a forwarded message and check if the indication is shown.

### Screenshots

Forwarded-messages21


feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/ConversationChatListScreen.kt
```diff
@@ -117,6 +117,7 @@ internal fun ConversationChatListScreen(
 
     LaunchedEffect(courseId, conversationId) {
         chatListState.scrollToItem(0)
+        viewModel.onCloseThread = null
```
FelberMartin, line 120: Is this one of the bugfixes mentioned in the PR description?

feature/metis/conversation/src/test/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ConversationMessagesUITest.kt
```diff
@@ -138,13 +162,72 @@ class ConversationMessagesUITest : BaseChatUITest() {
...
+            forwardedPosts = listOf(null)
+        )
+
+        composeTestRule.onNodeWithTag(TEST_TAG_DELETED_FORWARDED_POST, useUnmergedTree = true)
+            .assertExists()
+    }
+
+    // ##########################################################################################
+
+    private fun testForwardedMessageInVisibility() {
```
FelberMartin, line 213: Rename to `testForwardedMessageInvisibility` (with a lowercase "v"), otherwise its hard to read properly :D

feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/chatlist/ChatListItem.kt
```diff
@@ -1,17 +1,87 @@
...
+    sealed class PostItem : ChatListItem() {
+        abstract val post: IBasePost
+        val key: Any
+            get() = post.key
+
+        abstract fun copy(post: IBasePost = this.post): PostItem
+
+        interface ForwardedMessage {
+            val forwardedPosts: List<IBasePost?>
+            val courseId: Long
```
FelberMartin, line 25: The `courseId` feels kinda off here, is it only used for creating the inAppLink? If so, I'd suggest to create a new `CommunicationDeepLinks.ToConversationCourseAgnostic`, similar to eg `ToLectureCourseAgnostic`.  

Also, would it then make sense to remove the `interface ForwardedMessage` and its implementations, and instead move `forwardedPosts` to the PostItem and leave it empty for posts without forwarded messages? But maybe I missed something, please let me know :D

FelberMartin (changes requested):
Thank you for working on this, this forwarding thingy really turned out to be much more complex than anticipated :D

Test the functionality and works as described, just a few code comments :)

feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/saved_posts/SavedPostItem.kt
```diff
@@ -60,9 +61,12 @@ fun SavedPostItem(
         trailingCardContent = trailingCardContent
     ) {
 
+        // TODO: A ChatListItem.PostItem should be passed here to support forwarded messages
+        // https://github.com/ls1intum/artemis-android/issues/459
         PostItemMainContent(
-            modifier = Modifier,
+            modifier = Modifier.wrapContentHeight(unbounded = true),
```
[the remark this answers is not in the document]
julian-wls: When not using `wrapContentHeight(unbounded = true)` I had the following problem: The Post Main content was always cut off and the ignored the height set by the parent row. I think this was also related to the fact that the quote border box had its height set to `IntrinsicSize.Min`.

feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/ConversationChatListScreen.kt
julian-wls: Yes, I added this line to fix a problem that occurred when deleting a post.

feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/chatlist/ChatListItem.kt
julian-wls: Could you elaborate more on the `CommunicationDeepLinks.ToConversationCourseAgnostic`? I'm not 100% sure, but do we have to extend the existing deeplinks for that? I tried some approaches, but I was not successful yet...

I think I would leave the `interface ForwardedMessage` in place as it is really easy to check if  a ChatListItem is a forwarded message that way. I actually used it quite often.

FelberMartin: > Could you elaborate more on the CommunicationDeepLinks.ToConversationCourseAgnostic? I'm not 100% sure, but do we have to extend the existing deeplinks for that? I tried some approaches, but I was not successful yet...

Ah I see, the problem is that the courseId is a required argument for the `CourseUiScreen` route. Hm, I did not think of that... Then I think its fine for now, maybe I'll tackle this once I implement `ArtemisContext` to contain a courseId. 

> I think I would leave the interface ForwardedMessage in place as it is really easy to check if a ChatListItem is a forwarded message that way. I actually used it quite often.

As said, I think its fine for now, but in case I implement the latter `ArtemisContext`, would a simple function `PostItem.hasForwardedMessages() = forwardedMessages.isNotEmpty()` be sufficient?

FelberMartin (approved):
Lgtm now.

We can remove the `ToConversationCourseAgnostic` then :)

feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/chatlist/ChatListItem.kt
julian-wls: Yes, I think so

eylulnc (approved):
In code wise I have nothing to say 🙈 
Functionality looks good. This is a good update to align with the web and will improve lot user experience.

PR #456, opened 2025-03-01
`Chore`: Adapt to server api change

FelberMartin:
### Problem Description
https://github.com/ls1intum/Artemis/pull/10416 introduced breaking changes to the server APIs. This PR updates the API calls to cope with these changes

### Changes
 
- Introduced central `Api` class to manage all the high-level routes
- Use this class in all api calls

### Steps for testing
1. Go to a server with the new API routes enabled (you can see whether these are enabled in the network inspector. Eg the `time` request has route `api/core/public/time` after the changes and `api/public/time` prior to the server api changes)
2. Log in to such a server with Android
3. See that everthing network related works as before


julian-wls (changes requested):
Looks good overall,
I noticed a problem with the Pdf Viewer. If you click on any pdf (chat/lecture) the file can't be found and the viewer closes.

FelberMartin:
Fixed the member and user search, for the pdfs I created an issue #458, because it is not clear yet whether the root cause of this problem should be fixed by the clients or the server.

Regarding the course icons, that is indeed really weird. I could also observe weird behaviour regarding the course icons on the webapp. Sometimes some of the icons do not load, for others there the server responds with a 404 and sometimes when I just call the icon api url, the returned image cannot be displayed because "The image contains errors".

So I think I'll add another item to the above issue, to check whether this issue is resolved before we release the new app version, if that's fine :)

julian-wls (approved):
Member search is working now, didn't notice any other issues.
> Fixed the member and user search, for the pdfs I created an issue https://github.com/ls1intum/artemis-android/issues/458, because it is not clear yet whether the root cause of this problem should be fixed by the clients or the server.

I think this should be fine then, regarding the course icons I also agree with you.

Code lgtm.

PR #461, opened 2025-03-05
`Feature`: Add Grouping to Exercise and Lecture List

eylulnc:
### Problem Description
Currently, lectures and exercises on Android are only grouped by weeks (GroupedByWeek). This differs from the iOS approach, which divides items into Past, Current, Future, No Date, and Due Soon sections. This leads to inconsistent grouping between platforms and a suboptimal user experience.

### Changes
 
New grouping logic:
Introduced a TimeFrame<T> sealed class (past/current/future/noDate/dueSoon).
Added an extension function to map Lecture / Exercise lists into these group types based on startDate/endDate logic, mirroring iOS.
New List Screen:
A Compose screen that displays these groups in collapsible headers.
Replaced old “weekly grouping” code:
Removed or deprecated GroupedByWeek<T> usage in the main lecture/exercise screens.
Updated CourseViewModel to produce DataState<List<TimeFrame<Lecture|Exercise>>> instead of the old weeklyLecturesDataState and weeklyExercisesDataState.

### Steps for testing
Open a Course in the Android app that has several lectures and exercises with varying start/end dates 

Navigate to the Lectures tab:
Expect to see collapsible sections labeled “Past (n)”, “Current (n)”, “Future (n)”, or “No date (n)” if no start/end is defined.
Toggling a section header expands or collapses its items.

Navigate to the Exercises tab:
Exercises are shown similarly in collapsible sections.
If “Due soon” is enabled, you should see an extra “Due soon (n)” group for exercises with dueDate < 3 days away.

### Screenshots
Screenshot 2025-03-06 at 00 24 53


eylulnc:
Do you have any suggestion for fixing the E2E test?

feature/course-view/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/courseview/ui/TimeFrameItemsLazyColumn.kt
```diff
@@ -0,0 +1,158 @@
...
+@Composable
+private fun <T> TimeFrameItemsSectionHeader(
+    modifier: Modifier,
+    group: TimeFrame<T>,
+    expanded: Boolean
+) {
+    val size = group.items.size
+    val groupTitle = when (group) {
+        is TimeFrame.Past<*> -> {
+            "Past (${size})"
```
FelberMartin, line 101: We should introduce string resources for this

feature/course-view/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/courseview/ui/TimeFrameItemsLazyColumn.kt
```diff
@@ -0,0 +1,158 @@
...
+        is TimeFrame.Future -> {
+            "Future (${size})"
+        }
+        is TimeFrame.NoDate -> {
+            "No date (${size})"
+        }
+        is TimeFrame.DueSoon -> {
+            "Due soon (${size})"
+        }
+    }
```
FelberMartin, line 115: `groupTitle += " ($size)"`

FelberMartin (changes requested):
Nice change, I like the alignment with iOS :)

Besides the code comments, one general comment:
I think we should also sort the items within a TimeFrame (I think this is also done so in iOS). Currently the items do not seem to follow a meaningful ordering:
image

FelberMartin:
> Do you have any suggestion for fixing the E2E test?

It seems like the exercises/lectures created by the E2E tests have the TimeFrameGroup "noDate", which are collapsed by default. So you first need to expand the lists in `displayExerciseTypeTestImpl` and `LectureListE2eTest` to be able to scroll to them :)

julian-wls (commented):
Nice changes, works as expected.

I agree with Martin, could we also actually show this timeframe in a small header like it is done on iOS?
I have one additional remark regarding the searching for an item:
I don't know if that's feasible in the scope of this PR, but I would find it nice if the sections are expanded when the result item is in a certain section, currently it looks like this when searching for an item:

image


I think the code seems fine for me.

eylulnc:
> Nice changes, works as expected.
> 
> I agree with Martin, could we also actually show this timeframe in a small header like it is done on iOS? I have one additional remark regarding the searching for an item: I don't know if that's feasible in the scope of this PR, but I would find it nice if the sections are expanded when the result item is in a certain section, currently it looks like this when searching for an item:
> 
> image
> 
> I think the code seems fine for me.

I will take a look at it as a follow up

eylulnc:
> It seems like the exercises/lectures created by the E2E tests have the TimeFrameGroup "noDate", which are collapsed by default. So you first need to expand the lists in `displayExerciseTypeTestImpl` and `LectureListE2eTest` to be able to scroll to them :)

Thanks 😊

FelberMartin (approved):
Code and functionality lgtm!

PR #464, opened 2025-03-07
`Feature`: Add native emoji picker

FelberMartin:
### Problem Description
Users were not able to directly add emojis to their posts, like they are used to from other messaging apps. For example WhatsApp or Slack both offer a emoji button that takes the user to a native in-app emoji picker.

### Changes
 
- Moved all emoji related code to new submodule `feature:metis:conversation:emoji-picker`
- Created custom emoji picker and removed dependecy to `emoji2` (for reasons, [see comment in `EmojiPicker.kt`](https://github.com/ls1intum/artemis-android/pull/464/files#diff-9b6b3cca0d5726d8b03c21735cf3df69b49e878e38a82ad6f86edd05be949017R46))
  - Display emojis by category
  - Save recently used emojis
  - Allow search for emoji by identifier
- Use this new picker for reactions and for adding emojis to posts (via a new button)

### Steps for testing

1. Go to any chat
2. Open the ReplyTextField
3. Add an emoji to your message using the newly added emoji button
4. Add a new reaction to a post via the emoji picker
5. Verify that the search in the emoji picker works as expected.

**Known issues**:
- The performance in the debug build is quite bad and you may experience framedrops while eg scrolling through the emoji list. This is (probably) caused by the huge number of composables being displayed by the LazyGrid and [performance problems in debug build](https://developer.android.com/develop/ui/compose/lists#measuring-performance). I tested the changes with the production build, and there everything was smooth.
- When entering the EmojiPicker from the ReplyTextField button, there is a quick "jump" of BottomSheet. I was not able to prevent this from happening and this is probably also related to [this commented issue](https://github.com/ls1intum/artemis-android/pull/464/files#diff-30826aa01431745b2806a2a33585f3e90974cd4075f886b5a478f6b586f40e1aR41)


eylulnc (commented):
Looks really good! 

One slight question for the search bar
Screenshot 2025-03-11 at 07 39 30
Screenshot 2025-03-11 at 07 39 37

I think its only matches from the start but not the text included.
If its to much effort I think its fine.

julian-wls (changes requested):
Awesome change :) this is a really great improvement for the app!

Tested it and it works as expected :+1: 
I have one general remark: 
Can we overwrite the searchBarBackgroudColor for the modal bottom sheet using `MaterialTheme.colorScheme.surfaceContainerHigh`. I think the default surfaceContainer does not provide enough contrast to the background.

And maybe we could consume this space and add contentPadding to the LazyVerticalGrid instead, what do you think:

image

FelberMartin:
I address both you comments, and improved the search :D 
image

julian-wls (approved):
Really nice, thanks for implementing my feedback.
Looks all nice and shiny now!
Code lgtm, can be merged :)

eylulnc (approved):
Thanks for making the changes looks amazing ☺️

PR #468, opened 2025-03-10
`Feature`: Allow users to forward messages

julian-wls:
### Problem Description

This PR adds the missing forwarding functionality to the Android app.

### Changes

The following changes are introduced:
- New post action for forwarding messages
- New post forwarding bottom sheet
- An alternative to the existing member selection has been added, that uses a drop down popup to allow member and channel selection (This could be used for other features, too. I tried to keep it general)
- The Post send worker has been adjusted to check for attached forwarded messages and create them if necessary
- added an alternative to the markdownTextField so it can be displayed in the bottom sheet.
- This PR includes some additional bug fixes and improvements

Note: The refactoring of the markdown textfield caused the unit tests to fail. I spend a lot of time trying to fix them, but did not find a solution. @FelberMartin do you have an idea why they are not finding the test tag anymore?

### Steps for testing

1. Long-click on any post that contains a message or a forwarded message and verify the forward option is shown
2. Verify it is not shown for posts that do not contain a message and have a deleted forwarded message.
3. Click on forward post and verify that the bottom sheet shows up.
4. Select a user and a channel and add a message to verify the text field is working as expected.
5. Send the message and head to the channel to verify the message appeared. Verify that if you selected a user, which a DM chat was not present a new chat has been created. Same applies to groups, when multiple users are selected.
6. Verify that Toasts are shown correctly.

Note: The Tagging dropdown menu does cover the keyboard now, I did not find a way to open it above the keyboard. The issue is related to the text formatting options being placed in a box now.


feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/post/util/ForwardMessageUseCase.kt
```diff
@@ -0,0 +1,173 @@
...
+    fun forwardPost(post: IBasePost, onComplete: (Boolean) -> Unit) {
+        viewModelScope.launch(coroutineContext) {
+            val (recipients, conversations, serverData) = combine(
+                recipients,
+                conversations,
+                serverConfigurationService.serverUrl,
+                accountService.authToken
+            ) { recipients, conversations, serverUrl, authToken ->
+                Triple(recipients, conversations, serverUrl to authToken)
+            }.first()
```
FelberMartin, line 96: I may be wrong, but we could just use `val recipients = recipients.first(); val conversations = conversations.first() ...`?

feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/post/util/ForwardMessageUseCase.kt
```diff
@@ -0,0 +1,173 @@
...
+
+            val forwardedSourcePostList = listOf(
+                ForwardedSourcePostContent(
+                    sourcePostId = post.serverPostId ?: -1,
```
FelberMartin, line 100: I am not sure, does it make sense to continue here with a `sourcePostId = -1`? Or would return make more sense?

feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/post/util/ForwardMessageUseCase.kt
```diff
@@ -0,0 +1,173 @@
...
+        }
+    }
+
+    private suspend fun forwardPostToRecipients(
+        recipients: List<MemberSelectionItem.Recipient>,
+        forwardedSourcePostList: List<ForwardedSourcePostContent>,
+        serverData: Pair<String, String>
+    ): Boolean {
+        val (serverUrl, authToken) = serverData
+        if (recipients.isNotEmpty()) {
```
FelberMartin, line 136: Early return in case of an empty list here :)

feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/reply/autocomplete/ReplyAutoCompletePopup.kt
```diff
@@ -58,11 +56,12 @@ internal fun ReplyAutoCompletePopup(
...
     ) {
-        val maxHeight = min(maxHeightFromScreen, AutoCompletionDialogMaxHeight)
-        val verticalPadding = Spacings.AutoCompletePopup.ContentVerticalPadding
+        val maxHeight = min(maxHeightFromScreen, Spacings.Popup.maxHeight)
+        val verticalPadding = Spacings.Popup.ContentVerticalPadding
         ReplyAutoCompletePopupBody(
             modifier = Modifier
-                .padding(vertical = verticalPadding)
+                .padding(vertical = verticalPadding, horizontal = 8.dp)
+                .dropShadowBelow()
```
FelberMartin, line 64: Is there a reason why we have two dropShadowBelows now? One here, one in `ReplyAutoCompletePopupBody`

feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/reply/ReplyTextField.kt
```diff
@@ -91,6 +93,10 @@ internal fun ReplyTextField(
     onFileSelected: (Uri) -> Unit,
     updateFailureState: (MetisModificationFailure?) -> Unit,
     conversationName: String,
+    surfaceShape: Shape = MaterialTheme.shapes.large.copy(
```
FelberMartin, line 96: A comment why we need this as a parameter would be nice :)

feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/thread/MetisThreadUi.kt
```diff
@@ -333,10 +343,12 @@ private fun PostAndRepliesList(
         state = state
     ) {
         item {
-            val postActions = rememberPostActions(post)
+            val chatListItem = chatListContextItem ?: ChatListItem.PostItem.ThreadItem.ContextItem.ContextPost(post)
             val linkPreviews by remember(post.content) {
                 generateLinkPreviews(post.content.orEmpty())
             }.collectAsState()
+            val displayForwardBottomSheet = remember(post) { mutableStateOf(false) }
+            val postActions = rememberPostActions(chatListItem, displayForwardBottomSheet)
```
FelberMartin, line 351: Same as in MetisChatList, can we use the `by` operator?

feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/thread/MetisThreadUi.kt
```diff
@@ -374,12 +389,13 @@ private fun PostAndRepliesList(
...
             key = { index, post -> post.clientPostId ?: index }
         ) { index, answerPost ->
-            val postActions = rememberPostActions(answerPost)
             val answerChatListItem = answerChatListItemState(answerPost).collectAsState()
+            val chatListAnswerItem = answerChatListItem.value ?: ChatListItem.PostItem.ThreadItem.Answer.AnswerPost(answerPost)
             val answerPostLinkPreviews by remember(answerPost.content) {
                 generateLinkPreviews(answerPost.content.orEmpty())
             }.collectAsState()
-
+            val displayForwardBottomSheet = remember(answerPost) { mutableStateOf(false) }
```
FelberMartin, line 397: and here aswell?

feature/metis/manage-conversations/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/manageconversations/ui/conversation/member_selection/MemberSelectionBaseViewModel.kt
```diff
@@ -108,25 +121,87 @@ internal abstract class MemberSelectionBaseViewModel(
     }
         .stateIn(viewModelScope + coroutineContext, SharingStarted.Eagerly)
 
+    private val conversationsFromServer: StateFlow<DataState<List<Conversation>>> = combine(
+        query,
```
FelberMartin, line 125: Currently we load _all_ conversations from the server every time the query changes. I think we should load allConversation once, and then just update the displayed ones based on the query. Also, the query for conversations is not debounced atm :)

feature/metis/manage-conversations/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/manageconversations/ui/conversation/member_selection/MemberSelectionDropdownPopup.kt
```diff
@@ -0,0 +1,254 @@
...
+            // Only update when it actually changes to avoid unnecessary recompositions and popup shaking
+            if (isPopupExpanded != newExpanded) {
+                isPopupExpanded = newExpanded
+            }
+        },
+        hint = stringResource(id = R.string.conversation_member_or_channel_selection_address_hint),
+        focusRequester = focusRequester
+    )
+
+    if (isPopupExpanded && suggestions.isNotEmpty()) {
```
FelberMartin, line 94: early return if we do not show the popup

feature/metis/manage-conversations/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/manageconversations/ui/conversation/member_selection/MemberSelectionDropdownPopup.kt
```diff
@@ -0,0 +1,254 @@
...
+        val screenHeight = LocalConfiguration.current.screenHeightDp.dp
+        val maxHeightFromScreen = screenHeight * 0.4f
```
FelberMartin, line 96: magic number

FelberMartin (changes requested):
Really like the introduction of the ForwardMessagesUseCase and the generic implementation of the member selection, good job :D 

Besides the code comments (I know there are a lot, sorry), I have two general remarks:
1. Sometimes when typing into the search bar in the forward BottomSheet, I got serialization errors, I think caused by bad requests. I was able to reproduce this quite consistently on both TS0 and TS2:
grafik

2. When opening the forward BottomSheet multiple times in a chat (for different posts), the "draft" data does not reset, i.e. when selecting recipients and adding the post content, then closing the BottomSheet, and open it for another post, this data persists. This behaviour may be caused by the saved state handle usage in the `MemberSelectionBaseViewModel`, but I think we can discuss this as part of a separate ticket, this PR is big enough :D

feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/ui/post/post_actions/PostForwardBottomSheet.kt
```diff
@@ -0,0 +1,159 @@
...
+    forwardMessageUseCase: ForwardMessageUseCase,
+    onDismissRequest: (Boolean) -> Unit
+) {
+    val context = LocalContext.current
+
+    val onSendMessage = forwardMessageUseCase::forwardPost
+    val post = chatListItem.post
+    val postToForward =
+        if (chatListItem is ChatListItem.PostItem.ForwardedMessage) chatListItem.forwardedPosts.firstOrNull()
+            ?: post else post
```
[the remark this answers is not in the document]
julian-wls: I just checked: I think this applies to posts that have a content, otherwise it's different. I'll change it to represent this logic.

feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/service/impl/CreatePostServiceImpl.kt
```diff
@@ -22,6 +23,8 @@ internal class CreatePostServiceImpl(private val context: Context) : CreatePostS
         courseId: Long,
         conversationId: Long,
         content: String,
+        hasForwardedMessage: Boolean,
+        forwardedSourcePostList: List<ForwardedSourcePostContent>?,
```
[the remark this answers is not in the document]
julian-wls: I really don't know why, but for some reason when removing the `hasForwardedMessage` flag the forwarded messages are not created by the server. I tested it and all functions are called correctly. I would suggest to leave the parameter in place.

julian-wls:
Thanks for your valuable comments and suggestions. I tried to fix all of the issues mentioned above. The serialization issue has also been resolved now. Regarding the saved draft data I'm not sure if it is a feature or a bug :) I kept it for now.

feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/service/impl/CreatePostServiceImpl.kt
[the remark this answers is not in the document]
FelberMartin: But can't we then just use `hasForwardedMessages = forwaredMessages?.orEmpty.isNotEmpty()` in `SendConversationPostWorker.uploadPost()` instead of passing this boolean all the way down?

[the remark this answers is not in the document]
julian-wls: I tried, but it never worked for me.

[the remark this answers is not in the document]
FelberMartin: That sounds strange. I mean I did not try any potential edge cases, but I just check with the following code change, and the post was forwarded correctly and sending normal posts worked as well on TS0
image

FelberMartin (changes requested):
Thanks for addressing the comments, I found one more UI bug while testing:

When forwarding a message with a long post body (eg with an image), the layout looks messed up when opening the keyboard:
image

FelberMartin:
> Note: The Tagging dropdown menu does cover the keyboard now, I did not find a way to open it above the keyboard. The issue is related to the text formatting options being placed in a box now.

Regarding the placement of the popup dialog, I think this is actually fixable. You can have a look at how it is done for the `ReplyAutoCompletePopup` in `ReplyTextField:220` (wrapping the popup inside a `Box` and getting the max available size from there).

feature/metis/shared/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/shared/ui/member_selection/MemberSelectionDropdownPopup.kt
```diff
@@ -0,0 +1,258 @@
...
+import de.tum.informatics.www1.artemis.native_app.feature.metis.shared.ui.profile_picture.ProfilePicture
+import de.tum.informatics.www1.artemis.native_app.feature.metis.shared.ui.profile_picture.ProfilePictureData
+
+private const val POPUP_SCREEN_HEIGHT_FACTOR = 0.4f
+
+@Composable
+fun MemberSelectionDropdownPopup(
+    modifier: Modifier,
+    query: String,
+    focusRequester: FocusRequester,
```
FelberMartin, line 61: This composable does not have a root layout. Either provide a root layout, or add ColumnScope to the composable definition. This may also relate to the positioning issue of the popup

julian-wls:
> > Note: The Tagging dropdown menu does cover the keyboard now, I did not find a way to open it above the keyboard. The issue is related to the text formatting options being placed in a box now.
> 
> Regarding the placement of the popup dialog, I think this is actually fixable. You can have a look at how it is done for the `ReplyAutoCompletePopup` in `ReplyTextField:220` (wrapping the popup inside a `Box` and getting the max available size from there).

Oh :) I was actually referring to the tagging dropdown menu that is opened from the text field options. I don't think it is a problem if the member selection popup covers the keyboard a little bit. I added a max height for this.

FelberMartin:
> I don't think it is a problem if the member selection popup covers the keyboard a little bit. I added a max height for this.

It's just a minor detail, yep, but I think it should be not that hard to fix. This way, we could maybe also get rid of the `POPUP_SCREEN_HEIGHT_FACTOR = 0.4f` in `MemberSelctionDropdownPopup`. I don't really get where this comes from :)

julian-wls:
> It's just a minor detail, yep, but I think it should be not that hard to fix. This way, we could maybe also get rid of the `POPUP_SCREEN_HEIGHT_FACTOR = 0.4f` in `MemberSelctionDropdownPopup`. I don't really get where this comes from :)

I'll take a look :+1:

feature/metis/conversation/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/service/impl/CreatePostServiceImpl.kt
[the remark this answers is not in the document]
julian-wls: For me the following happens: The post get send correctly, the forwarded post gets created correctly and is sent to the server, but as soon as I navigate to the chat I can see the created post, but not the forwarded post underneath.

FelberMartin (commented):
The dialog changes lgtm, I did not experience any clipping problems anymore 👌

eylulnc (approved):
Looks really good you made a lot of effort ✅

PR #480, opened 2025-03-13
`Chore`: Move savedPosts feature to sub-module

FelberMartin:
### Problem Description
The feature:metis:conversation module is quite packed and cluttered. With this PR I try to extract some features out of it

### Changes
 
- Moved extractable code related to saved posts feature to own sub-module
- Created feature:metis:conversation:shared on the way, because otherwise we would have introduced circular dependencies

### Steps for testing
Ensure that everything related to saved posts works as expected


julian-wls (approved):
Changes seem reasonable to me :+1: 
Tested and works like before.

PR #488, opened 2025-03-15
`Improvement`: Redesign channel creation

julian-wls:
### Problem Description

There are some inconsistencies in the create channel view. The descriptions mention selective and public channels without the option being available in the first place.

### Changes

This PR makes some improvements to the CreateChannelScreen to the usability of the channel creation.

Note: After making the changes I noticed that there seems to be no appropriate way to test the state of segmented buttons, which is why all tests are currently failing in CreateChannelE2ETest. Do you have any suggestions?

### Steps for testing

Make sure you can create channels like before and make sure everything works as expected.

### DIFF
--- a/feature/metis/manage-conversations/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/manageconversations/ui/conversation/create_channel/CreateChannelScreen.kt
+++ b/feature/metis/manage-conversations/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/manageconversations/ui/conversation/create_channel/CreateChannelScreen.kt
@@ -89,9 +89,9 @@ internal fun CreateChannelScreen(
     val isNameIllegal by viewModel.isNameIllegal.collectAsState()
     val isDescriptionIllegal by viewModel.isDescriptionIllegal.collectAsState()
 
-    val isPublic by viewModel.isPublic.collectAsState()
+    val visibility by viewModel.visibility.collectAsState()
     val isAnnouncement by viewModel.isAnnouncement.collectAsState()
-    val isCourseWide by viewModel.isCourseWide.collectAsState()
+    val scope by viewModel.scope.collectAsState()
 
     val canCreate by viewModel.canCreate.collectAsState()
 
@@ -137,7 +137,7 @@ internal fun CreateChannelScreen(
                     } else R.string.create_channel_text_field_name_hint
                 ),
                 leadingIcon = {
-                    if (isPublic) {
+                    if (visibility == CreateChannelViewModel.Visibility.PUBLIC) {
                         Icon(
                             modifier = Modifier,
                             imageVector = Icons.Default.Numbers,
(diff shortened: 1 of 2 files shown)


FelberMartin (commented):
Nice change, code lgtm :D

Regarding testing, one idea might be to add the `selected` boolean as part of the test tag.

feature/metis/manage-conversations/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/manageconversations/ui/conversation/create_channel/CreateChannelScreen.kt
```diff
@@ -211,6 +205,142 @@ internal fun CreateChannelScreen(
...
+                        onClick = {
+                            selectedOption = index
+                            onCheckedChange(selectedOption == targetOption)
+                        },
+                        shape = SegmentedButtonDefaults.itemShape(index, buttonLabelOption.size),
+                        icon = {
+                            SegmentedButtonDefaults.Icon(
+                                active = selected,
+                                activeContent = {
+                                    Icon(
```
FelberMartin, line 318: I noticed that the check icon animation is a little bit juddery now, but I also noticed the padding issue without adding this icon. Also after some research I did not find a good solution for this :/

julian-wls: Yea, I was also debugging this way too long, but without success. I'm disappointed I had to go with this solution...

FelberMartin: Damn, maybe add a comment about this problem to the code, but as I think we should not invest too much time into this detail, I'm fine with merging this this way :)

feature/metis/manage-conversations/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/manageconversations/ui/conversation/create_channel/CreateChannelScreen.kt
```diff
@@ -211,6 +205,142 @@ internal fun CreateChannelScreen(
...
+@Composable
+private fun SegmentedSelection(
+    modifier: Modifier,
+    title: String,
+    description: String,
+    buttonLabelOption: List<String>,
+    onCheckedChange: (Boolean) -> Unit,
+    testTag: String = ""
+) {
+    val targetOption = 0
```
FelberMartin, line 275: Instead of defaulting to the first option as being selected, this state should come as a parameter from the viewmodel. Currently we have inconsistent state: The UI shows "course-wide" as selected, while the viewmodel initializes with "selective".

FelberMartin (approved):
One more code comment, apart from that good to go from my side with the discussed testing changes

julian-wls:
Thanks for clarifying. Makes a lot more sense now!

FelberMartin (approved):
Awesome, thank you for tackling this!

PR #500, opened 2025-03-21
`Chore`: Replace .sp usage with MaterialTheme.typography

eylulnc:
### Problem Description
There are multiple usage of .sp font sizes across the app, where MaterialTheme.typography. should be used instead. Some of the usages of .sp do make sense, but eg in the AccountUI there are usages that should be replaced.

### Changes
 
Every possible sp usage changed with the closest MaterialTheme.typography match

- DataStateUI
- InfoMessageCard
- CourseItem
- AccountUI
- LoginUI

### Steps for testing
Please check the mentioned screens and verify everything look as before.

### Screenshots

An example:
Screenshot 2025-03-21 at 11 08 54


FelberMartin (approved):
This is a very neat PR for being the 500th in this repo 🎉 

Changes lgtm :D

PR #505, opened 2025-03-24
`Bugfix`: Back Navigation Work Around

eylulnc:
### Problem Description
As discussed in [#354](https://github.com/your-repo-link/issues/354), navigating from a channel to a lecture or exercise results in inconsistent back navigation behavior:

The Android system back button works as expected.

The top app bar back button does not trigger the expected back navigation.

This issue occurs because the NavigationBackButton previously relied solely on the provided onNavigateBack callback, which might be null or not behave consistently depending on the navigation flow.

### Changes
 
If onNavigateBack is not provided, the function now uses the default LocalOnBackPressedDispatcherOwner to trigger a system back event.

This ensures that the back button on the top app bar behaves similarly to the system back button.

### Steps for testing

1. Go to a channel that has a corresponding exercise/lecture
2. Click on the info button
3. Click on Go to exercise/lecture
4. Try to navigate back using both the Android back button and the top bar back button
5. Behaviour should be same

⚠️ Note: This is a Workaround
This solution is not a mandatory or permanent fix, but a workaround to unify navigation behavior in the short term without refactoring the navigation stack. It allows for consistent UX while keeping the codebase stable. A more robust solution could involve handling the back stack explicitly with navigation state.

Update!!!
Please check for the following screens if the navigation back is work as is used to be.

- FaqDetail
- Lecture 
- Exercise 
- RegisterForCourse
- Settings


FelberMartin (changes requested):
Awesome change, I did really not expect this to be just these few lines of code :D 

From my point of view this is even a perfect permanent fix, I dont see a reason why we would prefer handling the back stack explicitly instead of this solution.

My only request would be to make even more use of this change :D We could also apply this change to the other "simple" usages of `NavigationBackButton`. With "simple" I mean where the callback passed to `NavigationBackButton` is essentially just `navController::navigateUp` passed all the way down from the `RootNavGraph`. We can then also remove all the now unused onNavigateBack callbacks :D

julian-wls (approved):
Really nice change, works as expected.
I agree with @FelberMartin. This can be a permanent fix.

FelberMartin:
> Do we want to do it here or in a follow up?

Doing it here would be awesome :D

eylulnc:
@FelberMartin @julian-wls  I have update the code if you think I have missed a navigation please infor me but I belive it should be good now.

I have also update the description so you can check what else should be tested.

FelberMartin (commented):
Awesome, navigation still works as expected.

I found a few more screens that could make use of this:
- AccountSettingsScreen
- ViewQuizResultScreen
- CourseScaffold
- ? It may even be possible to use this in SinglePageConversationBody, but lets investigate this in another PR

eylulnc:
> Awesome, navigation still works as expected.
> 
> 
> 
> I found a few more screens that could make use of this:
> 
> - AccountSettingsScreen
> 
> - ViewQuizResultScreen
> 
> - CourseScaffold
> 
> - ? It may even be possible to use this in SinglePageConversationBody, but lets investigate this in another PR

I will take a look at it for the first two. Course is not possible, if I change it it breaks the logic for example

- Open course 
- Click lecture tab
- Use app back button to go back
- It will go to exercise tab

But our current implementation is correct I should go back to course list.

FelberMartin:
> I will take a look at it for the first two. Course is not possible, if I change it, it breaks the logic for example
> 
>     * Open course
> 
>     * Click lecture tab
> 
>     * Use app back button to go back
> 
>     * It will go to exercise tab
> 
> 
> But our current implementation is correct I should go back to course list.

Ohh, thats right, good catch! Yeah then lets leave the CourseScaffold for now :D

PR #506, opened 2025-03-24
`Chore`: Use Stateflow instead of cold Flow in ArtemisContextProvider

FelberMartin:
### Problem Description
The ArtemisContext is much better represented by a hot StateFlow. This way, we do not always have to initialize a hot flow with `ArtemisContext.Empty`, resulting in unnecassary reloads and API calls. For example the following error log was often produced for loading profile pictures, because of this problem: `Error loading image: Failed to connect to localhost/127.0.0.1:80`

### Changes
 
Use hot stateFlow with `MainScope()` instead of cold flow for providing the ArtemisContext.

### Steps for testing
See that stuff related to the changed files still works and eg the before-mentioned error log is no longer produced.


julian-wls (approved):
Couldn't reproduce the error in the logs any longer.
Code lgtm :+1:

PR #509, opened 2025-03-25
`Chore`: Switch to artemis.tum.de instance

FelberMartin:
### Problem Description
We should use the new tum artemis url now

### Changes
 
- Using new tum artemis url
- Replace legacy url on the fly if previously configured

### Steps for testing
- on develop: Login to the TUM instance
- checkout this PR: 
  - see that user is not logged out
  - see that url is now artemis.tum.de (eg in settings)
  - see that api calls still work


julian-wls (approved):
Works as expected.
Code lgtm!

Issue #523, opened 2025-03-28
Update KIT Artemis URL


julian-wls:
### Problem Description

Once [this PR](https://github.com/ls1intum/Artemis/pull/10595) is merged on the Artemis repo the url for the artemis instance of KIT is added.

This should be reflected on the Android app, too.

PR #525, opened 2025-03-31
`Chore`: Update KIT url

FelberMartin:
### Problem Description
See #523. Update the KIT artemis instance URL


julian-wls (approved):
Code lgtm. Can confirm, that this is the correct url.

PR #537, opened 2025-04-03
`Feature`: Feature Toggle Base Mechanism

eylulnc:
### Problem Description

The Android app previously lacked support for a **feature toggle mechanism**, which is useful for dynamically enabling or disabling functionality via the backend.

Available features are listed in the `/management/info` endpoint — the same endpoint used for force update checks. Therefore, the logic for **force update** and **feature availability** has been **merged** into a single flow.

Additionally, the old system waited 2 days before rechecking for updates. Since **feature availability** needs to be checked more frequently, this delay has been removed.

### Changes
 

- The update check now parses feature toggles from the /management/info response and makes them available through FeatureAvailability.
- Update checks are restarted when the server URL changes, ensuring that the latest app version and feature availability are always fetched in time — without waiting for a fixed delay. 
- The check runs only when the app is in RESUMED state and isLoggedIn, reducing unnecessary processing and ensuring updates appear promptly.
- Unit Test Extended for Feature Toggles

### Steps for testing
I have added the same step for force update please check that the logic works after user logged in. (In ios this logic trigger after logged in  check)

**Case 1: No Update Required**
The update screen should NOT be displayed if the server version matches the local app version.

- Ensure the server reports a version equal to the current app version.
- Launch the app.
- ( Log in  if you are not)
- Verify that no update prompt appears.

**Case 2: Update Required**
The update screen should be displayed if the server requires a newer version, and clicking the "Download" button should navigate to the Google Play Store.

- Locally modify the server response to return a higher version than the current app version. 
- Launch the app.
- ( Log in  if you are not)
- Verify that the UpdateScreen appears.
- Click the "Download" button.
- Confirm that the button redirects to the Google Play Store.
- If the app cannot reach the server, it should not force an update.

**Case 3: Feature Toggle**

- Observe the logs – every 60 seconds, the update and feature information should be fetched and printed.
- Confirm that feature availability is correctly set in logs.
- These logs will be removed after testing.

### Screenshots
Screenshot 2025-04-03 at 17 17 55

### DIFF
--- a/app/src/main/java/de/tum/informatics/www1/artemis/native_app/android/ui/MainActivity.kt
+++ b/app/src/main/java/de/tum/informatics/www1/artemis/native_app/android/ui/MainActivity.kt
@@ -66,6 +66,7 @@ import kotlinx.coroutines.launch
 import kotlinx.coroutines.runBlocking
 import org.koin.android.ext.android.get
 import org.koin.compose.koinInject
+import kotlin.time.Duration.Companion.seconds
 
 /**
  * Main and only activity used in the android app.
@@ -240,11 +241,9 @@ class MainActivity : AppCompatActivity(),
 
             LaunchedEffect(serverUrl, lifecycleOwner) {
                 lifecycleOwner.repeatOnLifecycle(Lifecycle.State.RESUMED) {
-                    if(isLoggedIn) {
-                        while (true) {
-                            updateRepository.triggerUpdateCheck()
-                            delay(60_000L)
-                        }
+                    while (isLoggedIn) {
+                        updateRepository.triggerUpdateCheck()
+                        delay(60.seconds)
                     }
                 }
             }
--- a/app/src/main/java/de/tum/informatics/www1/artemis/native_app/android/ui/MainActivity.kt
+++ b/app/src/main/java/de/tum/informatics/www1/artemis/native_app/android/ui/MainActivity.kt
@@ -13,13 +13,17 @@ import androidx.compose.material3.windowsizeclass.calculateWindowSizeClass
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.res.stringResource
 import androidx.core.net.toUri
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.compose.LocalLifecycleOwner
 import androidx.lifecycle.lifecycleScope
+import androidx.lifecycle.repeatOnLifecycle
 import androidx.navigation.NavHostController
 import androidx.navigation.compose.NavHost
 import androidx.navigation.compose.rememberNavController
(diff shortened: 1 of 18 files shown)


feature/force-update/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/force_update/repository/UpdateUtil.kt
```diff
@@ -1,53 +1,61 @@
...
+import de.tum.informatics.www1.artemis.native_app.feature.force_update.service.UpdateServiceResult
+
+private const val CHECK_INTERVAL_MS = 60_000L // 60 seconds
 
 object UpdateUtil {
 
-    /**
-     * Determines if it's time to check for an update (every 2 days).
-     */
     fun isTimeToCheckUpdate(lastCheckTime: Long, now: Long): Boolean {
```
FelberMartin, line 13: This function is now only used in the tests I think. Do we still need it?

app/src/main/java/de/tum/informatics/www1/artemis/native_app/android/ui/MainActivity.kt
```diff
@@ -225,17 +230,34 @@ class MainActivity : AppCompatActivity(),
...
+                .map { it is AccountService.AuthenticationData.LoggedIn }
+                .collectAsState(initial = false)
+            val serverUrl by serverConfigurationService.serverUrl.collectAsState(initial = "")
+
+            LaunchedEffect(serverUrl, lifecycleOwner) {
+                lifecycleOwner.repeatOnLifecycle(Lifecycle.State.RESUMED) {
+                    if(isLoggedIn) {
+                        while (true) {
+                            updateRepository.triggerUpdateCheck()
+                            delay(60_000L)
```
FelberMartin, line 246: There is another `delay` implementation taking a `Duration`. Imo its more readable: `delay(60.seconds)` :)

app/src/main/java/de/tum/informatics/www1/artemis/native_app/android/ui/MainActivity.kt
```diff
@@ -225,17 +230,34 @@ class MainActivity : AppCompatActivity(),
...
+            val updateResult by updateRepository.updateResultFlow.collectAsState()
+            val isLoggedIn by accountService.authenticationData
+                .map { it is AccountService.AuthenticationData.LoggedIn }
+                .collectAsState(initial = false)
+            val serverUrl by serverConfigurationService.serverUrl.collectAsState(initial = "")
+
+            LaunchedEffect(serverUrl, lifecycleOwner) {
+                lifecycleOwner.repeatOnLifecycle(Lifecycle.State.RESUMED) {
+                    if(isLoggedIn) {
+                        while (true) {
```
FelberMartin, line 244: I think this could lead to multiple `while(true)` loops running when we change the server url. I suggest to remove the if statement and change the loop to `while(isLoggedIn)`

FelberMartin (changes requested):
Just a few smaller code comments :)

feature/force-update/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/force_update/repository/UpdateUtil.kt
eylulnc: Sorry, my bad. I will remove it.

FelberMartin (approved):

julian-wls (approved):
Works as expected, nice addition :+1:

PR #541, opened 2025-04-08
`Chore`: Combine `ConversationAddMembersScreen` and `CreatePersonalConversationScreen`

### LINKED ISSUE #376: Combine `ConversationAddMembersScreen` and `CreatePersonalConversationScreen`
### Problem Description

Currently, there are two very similar composable functions `ConversationAddMembersScreen` and `CreatePersonalConversationScreen`, which both allow users to add other users to a conversation.

### Suggested changes

Both functions should be combined into one by extracting their common logic.

julian-wls:
### Problem Description

There were two very similar composables as described in #376.

### Changes
This PR introduces a new Composable `ConversationUserSelectionScreen`, which holds the common logic of both views. 
This PR fixes #376 

### Steps for testing
1. Go to the conversation overview.
2. Create a new chat.
3. Add members, click the create button and make sure the chat is correctly created.
4. Head over to the newly created chat and go to the conversation settings.
5. Click on add members.
6. Add new members and make sure everything works as expected.

### DIFF
--- a/feature/metis/manage-conversations/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/manageconversations/ui/common/ConversationUserSelectionScreen.kt
+++ b/feature/metis/manage-conversations/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/manageconversations/ui/common/ConversationUserSelectionScreen.kt
@@ -1,0 +1,117 @@
+package de.tum.informatics.www1.artemis.native_app.feature.metis.manageconversations.ui.common
+
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.WindowInsets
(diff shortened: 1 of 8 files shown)


FelberMartin (approved):
Nice change, code lgtm!

FelberMartin:
Two follow-ups that I noticed: 
1. We could display a small info box when creating a new chat, to inform the users that a DM will be created when only one user is selected, otherwise a group chat
2. When adding members to an existing conversation, I can still select users that are already members

julian-wls:
> * We could display a small info box when creating a new chat, to inform the users that a DM will be created when only one user is selected, otherwise a group chat
> * When adding members to an existing conversation, I can still select users that are already members

These are really good suggestions. As previously mentioned I am working on a redesign for the channel settings and could include these optimizations as part of this PR. What do you think?

FelberMartin:
Sounds good :D

PR #570, opened 2025-04-25
`Improvement`: Exercise view adjustments

julian-wls:
### Problem Description

The exercise view has no label to indicate the start of the problem statement and the problem statement comes with a gray background, which does not fit the rest of the view.

### Changes

This PR fixes the problems described above, by adding a new label and injecting a javascript code block to change the background color.

### Steps for testing

Make sure everything looks good and works as expected.


core/ui/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/core/ui/compose/ArtemisWebView.kt
```diff
@@ -137,6 +137,18 @@ private class ThemeClient(
             if (adjustHeightForContent) {
                 val delay = 750L
```
FelberMartin, line 139: Adding a comment explaining why this is necessary would be great :D

PR #577, opened 2025-05-02
`Bugfix`: Fix fileupload url

FelberMartin:
### Problem Description
After installing the Android app, the default serverUrl was set to the tum instance, but the slash was missing from the serverUrl. This caused a bug where file uploads incorrectly formatted the path to the uploaded file.

### Changes
 
This PR fixes this bug

### Steps for testing
- Uninstall and re-install the app
- Upload a file to a chat
- See that the uploaded image can be viewed


julian-wls (approved):
Works as expected. Code lgtm!

PR #596, opened 2025-05-09
`Communication`: Remove message background

eylulnc:
### Problem Description
Message background should be removed to align web.

### Changes
 
Removed the background from message.

### Steps for testing
Communication enabled course

1. Open communication channel, and wrote messages.
2. Verify that gray background is not there.
3. Special case background like (pin, saved) should be remain  the same

### Screenshots

Screenshot 2025-05-09 at 20 49 53


julian-wls:
I agree with @FelberMartin. Maybe there could be a compromise, such as making the background of the card lighter or adding a border, or adding more space between posts.... Just removing the background color does not feel right to me. I also agree that the design should be consistent across all platforms in the future.

eylulnc:
I have already removed in on iOS as well, Stephan ask this before "Why we have background here?", and in my discussion with Ramona she also agreed, that in the web we dont have the background, so lets remove it.

FelberMartin:
As commented, I am not a fan of this change, but I see that this decision comes from higher up, so whatever ..

FelberMartin (approved):

PR #603, opened 2025-05-12
`Bugfix`: Show correct user role badge after sending message

FelberMartin:
### Problem Description
There was an issue that the user role badge as a tutor sometimes was stuck on displaying "Student" even though the user is a "Tutor". After reloading the issue was gone. See https://github.com/ls1intum/Artemis/issues/10824

Additionally, the role badge shortly showed "Student" when sending a message, then it switched back to the correct author role.

### Changes
 
Fixed by removing the `remember` statement of the authorRole. Instead we handle remembering the author role for updates now in the database.
Additionally, I made the userRole nullable, so that we do not display any role badge when the role is null (instead of displaying "Student")

### Steps for testing
1. Login as a tutor (or instructor)
2. Go to any chat
3. Write a post
4. See that author role badge is shown correctly
5. Also test for creating answer posts and editing posts

### Screenshots

https://github.com/user-attachments/assets/fa0a1ffe-1543-4dad-8479-16bd65c8f557

### DIFF
--- a/feature/metis/conversation/shared/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/shared/ui/post/PostMainContent.kt
+++ b/feature/metis/conversation/shared/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/conversation/shared/ui/post/PostMainContent.kt
@@ -25,7 +25,6 @@ import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableLongStateOf
-import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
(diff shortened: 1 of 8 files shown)


julian-wls (commented):
After sending a message the app crashed with this stacktrace:
I'm not sure if this is related to this PR:
image

FelberMartin:
This is related to the PR, I forgot to mention this in the PR description, sorry. In Android Studio, go to "Build > Clean Build" and "Build > Rebuild". This should fix the error. To me it seems like the build files for the `MetisDao` are not rebuild by default

eylulnc:
I am still having the issue I tried to clean project, rebuild but could not do it

FelberMartin:
> I am still having the issue I tried to clean project, rebuild but could not do it

Hm that's weird. Another thing you could maybe try is to delete `MetisDao_Impl` manually in Android Studio, but tbh this is already what "clean project" should do...

eylulnc (commented):
Interestingly, I clean project and run it did not worked but after I  rebuild, delete the MetidDao_Impl and build it worked 😂

eylulnc (approved):
The user badge looks correctly, there is a small flicker with the last persons badge but I think that the part get re-trigger to find correct badge.

PR #606, opened 2025-05-13
`Feature`: View and download non-PDF lecture attachment units

### LINKED ISSUE #429: Lecture Attachments: Support more file types
### Problem Description

Currently, after #422 has been merged it is only possible to view and download pdf lecture attachments, but it is possible to upload various file types on the web app.

### Solution
 
A follow up PR should be created that introduces support for other file types than PDFs.

julian-wls:
### Problem Description

Currently, it is only possible to view pdfs when clicking on a lecture attachment unit. Other file types result in an error.

### Changes

This PR adds an image view to view lecture image attachment units and allows users to download unknown file types.
Therefore, ArtemisImageView has been created, which is used in the LinkBottomSheet, allowing users to view images in the bottom sheet and download and share the images.
For lecture attachment units that include other file types (not images or pdfs) a dialog is shown in which users can choose to download the file.
This PR closes #429.

### Steps for testing

1. Make sure everything related to pdfs does still works as intended (pdf sent to the chat and in lecture attachment units).
2. Go to a lecture that has image lecture attachment units and other, unknown file types as attachment units.
3. Click on these units.
4. Verify that images are shown in the image view.
5. Verify that a dialog shows up when clicking on a lecture attachment unit that includes an unknown file type (other than pdf or images).
6. Make sure the pdf lecture attachment units still work as expected.


eylulnc (changes requested):
The attachments opens as it should and  for unknown pop up available.

One small remark,
The image moves inside the bottom sheet ant make the buttom sheet hard to use, sometimes I can not make it extend or collapse is I move the image as in video
https://github.com/user-attachments/assets/27324478-7d9c-4783-9a79-3f4438aa143d

julian-wls:
Hi @eylulnc, thanks for your feedback. I completely redesigned the image view. Dragging the image is now only available after zooming in and the modal bottom sheet gestures are available as expected. Could you check again :)

FelberMartin (approved):
tested on with TS0, works as described. Code lgtm

eylulnc (approved):
Thanks for the changes, looking good

PR #612, opened 2025-05-15
`Improvement`: Change exercise and lecture section default collapse

FelberMartin:
### Problem Description
When opening a course that currently does not have any exercises in the "current" or "due soon" sections, the landing page looked quite empty. Instead we always want to show exercises, even when they are from a lower priority section, like eg "past". The same also applies to lectures.

### Changes
 
- Added new logic for setting the default section expansion state. The goal is to have at least 7 items on the page. As long as we have less, we keep expanding lower prio sections. 
  - **Section prio: current/dueSoon > Future > Past > NoDate**
- Scroll down to "current" or "dueSoon" when "future" is expanded (at the top of the list)
- Added unit tests

### Steps for testing
Go to different courses and verify in the exercise and lecture sections:
- There should always be exercises/lectures visible (if there are any) and the sections should be expanded based on the above mentioned section prios.

### Screenshots
image


eylulnc (approved):
All sections are expanded, and works as described

PR #613, opened 2025-05-15
`Feature`: Mute conversation via push notification action

FelberMartin:
### Changes
 
Added new action to mute a conversation when receiving a push notification for it.

### Steps for testing
1. Enable notifications for your device
2. Write a new message as another user to receive the push notification
3. On the notification, click "Mute"
4. See that the notification disappears
5. Navigate to the course's conversation list and confirm that the conversation is muted

### Screenshots
image


julian-wls (approved):
Works as expected, thanks for implementing this.
Maybe one remark: I'm not sure if 'Mute' makes it obvious to the user that this will only mute the conversation and not all notifications, but 'Mute Conversation' would probably be too long... What do you think @FelberMartin?

code lgtm :+1:

FelberMartin:
> Maybe one remark: I'm not sure if 'Mute' makes it obvious to the user that this will only mute the conversation and not all notifications, but 'Mute Conversation' would probably be too long

Exactly my thoughts as well :D 
I checked and for whatsapp its also just "Mute", so I decided to go with that

PR #629, opened 2025-05-22
`Bugfix`: Navigate to post after reply notification

### LINKED ISSUE #627: Notification navigation when clicking on thread reply
### Describe the bug
When clicking on a push notification of a thread reply, a thread view tries to load, but it just shows an error screen.

**To Reproduce**
1. Receive a push notification about a thread reply.
2. Click on it
3. See that the thread fails to load

**Expected behavior**
The thread should load.

FelberMartin:
### Problem Description
When clicking on a notification for a reply, the thread failed to load. This was for two main problems:
1. When storing the received notifications, we only stored the answerPostId for answer notifications. However, we need the id of the base post to navigate to the thread with the newly received answer.
2. When navigation to an older thread, the thread always failed to load. This is not directly related to the notification navigation, but could also be observed when receiving a notification about a new answer to an older post. This problem is also reproducible by saving an old post (go >20 posts up in the chat), and clicking on the post in the "Saved Messages" screen. Then the thread would also fail to load.

This PR closes #627.

### Changes
 
For fixing 1. :
- Store the basePost id as the `parentId` for answer notifications.
- Also, change the `metisTarget`s postId to the parentId. The postId is used to navigate to the thread.

For fixing 2. : 
- Wrap call to store fetched post in new `withTransaction` call
- Add `Log` calls to make future similar debugging easier

### Steps for testing
- Enable push notifications on the emulator
- With the webapp, write a new reply to an old post of the Android user (>20 posts above the newest post in the chat)
- Receive the push notification on Android
- Click on the notification
- See that the thread gets loaded successfully

### DIFF
--- a/core/ui/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/core/ui/common/DataStateUi.kt
+++ b/core/ui/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/core/ui/common/DataStateUi.kt
@@ -1,5 +1,6 @@
 package de.tum.informatics.www1.artemis.native_app.core.ui.common
 
+import android.util.Log
 import androidx.compose.animation.AnimatedContent
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.BoxScope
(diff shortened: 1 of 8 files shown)


eylulnc (approved):
On thread reply, notification received and navigated to thread ✅
Screenshot 2025-05-25 at 13 27 37

PR #635, opened 2025-05-27
`Improvement`: Show notification quick-reply texts

### LINKED ISSUE #216: Notification improvements

FelberMartin:
### Problem Description
When using the quick-reply feature to reply to a notification, the answer was not displayed in the notification (as expected eg from WhatsApp).
This PR closes #216 

### Steps for testing
1. Activate push notifications on Android
2. Receive a communication notification
3. Click "Reply" and write a reply to that notification
4. See that after sending the quick-reply, this reply text is also shown as a second part of the notification

### Screenshots
image


eylulnc (commented):
For the message on the top I annote the person than wrote the text, message did not display (both annotation and message) but for second one everything worked fine, I assume it is not related to this PR
Screenshot 2025-05-29 at 00 05 05

FelberMartin:
By "annotate", do you mean tagging the other user with `@test_user_1...`?

FelberMartin:
Got it, thanks. This might be related to the markdown handler, I'll have a look :D

PR #639, opened 2025-06-04
`Improvement`: Manage exercise buttons

eylulnc:
### Problem Description
The button structure of exercise was not align with Artemis Buttons. Also Start EXercise buttons shown even thought the due date is passed. Also to improve the UI we decided to use the exercise button as FAB.

### Changes
 
Display Start button if exercise due date is not passed.
Display Exercise Action Buttons as FAB
Use Artemis exercise button style
Problem statement is displayed only for programming exercise. (This is also discussed in iOS and implemented as well web view itself has a title called problem statement, so using a problem statement every exercise causes duplicate title, so we will only display it for the cases that title is not available)

### Steps for testing
Open Exercise see buttons display correctly as FAB.
Click on the button and check navigation works as before
See that if exercise due date is passed the button is not shown, also button is still shown for no due date exerises.
See that problem statement title is only visible for programming exercise. 
### Screenshots

Screenshot 2025-06-04 at 13 51 30


julian-wls (approved):
Tested and works as expected :+1: 
Code lgtm!

FelberMartin (approved):
Code lgtm

PR #658, opened 2025-06-26
`Improvement`: Use submission  to collect exercise results

eylulnc:
### Problem Description
In this [PR](https://github.com/ls1intum/Artemis/pull/10549) direct relationship between participation and result removed This causing some issue when displaying the graded exercise. This PR fixes that.

### DIFF
--- a/core/ui/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/core/ui/exercise/ParticipationStatusUi.kt
+++ b/core/ui/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/core/ui/exercise/ParticipationStatusUi.kt
@@ -39,8 +39,12 @@ fun ParticipationStatusUi(
 
     val participation =
         remember(exercise) { exercise.getSpecificStudentParticipation(showUngradedResults) }
-    val result = if (showUngradedResults) participation?.results.orEmpty()
-        .first() else participation?.results?.firstOrNull { it.rated == true }
+    val result = if (showUngradedResults) participation
+        ?.submissions
+        ?.firstOrNull()?.results.orEmpty()
+    else participation
+        ?.submissions
+        ?.firstOrNull()?.results?.filter { it.rated == true }?.sortedBy { it.completionDate }
 
     val length = if (isChip) TextLength.Short else TextLength.Full
     val textStyle =  if (isChip)  MaterialTheme.typography.labelMedium.copy(fontWeight = FontWeight.Bold, color = MaterialTheme.colorScheme.onSurface)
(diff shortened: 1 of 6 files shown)


core/ui/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/core/ui/exercise/ResultTemplateStatus.kt
```diff
@@ -130,11 +130,19 @@ fun computeTemplateStatus(
...
+        ?: participation.submissions
+            ?.flatMap { submission ->
+                submission.results
+                    ?.filter { it.rated == true }
+                    ?.map { it to (it.completionDate ?: Instant.fromEpochSeconds(0L)) }
+                    ?: emptyList()
+            }
+            ?.maxByOrNull { it.second }
+            ?.first
+        ?: participation.results.orEmpty()
```
FelberMartin, line 143: Could we also use the `findLatestRatedResult` method defined in the ExerciseViewModel here?

core/ui/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/core/ui/exercise/ParticipationStatusUi.kt
```diff
@@ -39,8 +39,12 @@ fun ParticipationStatusUi(
...
     val participation =
         remember(exercise) { exercise.getSpecificStudentParticipation(showUngradedResults) }
-    val result = if (showUngradedResults) participation?.results.orEmpty()
-        .first() else participation?.results?.firstOrNull { it.rated == true }
+    val result = if (showUngradedResults) participation
+        ?.submissions
+        ?.firstOrNull()?.results.orEmpty()
+    else participation
+        ?.submissions
+        ?.firstOrNull()?.results?.filter { it.rated == true }?.sortedBy { it.completionDate }
```
FelberMartin, line 47:
```suggestion
    var result = if (showUngradedResults) participation
        ?.submissions
        ?.firstOrNull()?.results.orEmpty()
   if (showUngradedResults) {
       result = result.filter { it.rated == true }?.sortedBy { it.completionDate }
   }
```

FelberMartin (commented):
Code lgtm, just two small comments

core/ui/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/core/ui/exercise/ParticipationStatusUi.kt
FelberMartin: Wait, I forgot to remove the first if statement. From my point of view this suggestion would help to remove the logic duplication, but if you strongly prefer the other option, we can also keep it as is :)

eylulnc: Nope good like this 😅, I have integrated your suggestion to my commit directly thanks for the suggestion

PR #665, opened 2025-07-12
`Bugfix`: Fix Archive/Favorite Toggle

### LINKED ISSUE #386: Archive and Favorite Status Toggles
There are inconsistencies with how archive and favorite statuses interact as mentioned [ in PR 374](https://github.com/ls1intum/artemis-android/pull/374), causing issues in scenarios where a channel should move between the "Archive" and "Favorite" sections.

Steps to Reproduce:

Scenario 1:

Archive a favorite chat (e.g., "paging").
Issue: The chat appears in both the "Archive" and "Favorite" sections instead of being removed from "Favorite."
Scenario 2:

Mark a previously archived channel (e.g., "test2") as a favorite.
Issue: The chat does not move from the "Archive" section to the "Favorite" section as expected.

Expected Behavior:
When a favorite chat is archived, the favorite status should be removed, and the chat should only appear in the "Archive" section.
When a previously archived chat is marked as favorite, it should move to the "Favorite" section and be removed from the "Archive" section.

Actual Behavior:
The chat remains visible in both sections or does not move between sections as expected. 

Possible Cause:
There might be a "double-toggle" in the code where two parallel network calls are executed, leading to inconsistent updates on the server.

eylulnc:
### Problem Description
The Archive and Favorite channel was working correct but it was not toggle the status of the other.
See the [issue](https://github.com/ls1intum/artemis-android/issues/386). This PR resolves that.

e.g if a channel is in favorite marks as archived than the favorite status should be toggled.

### Changes
 
Send a second request to toggle status.

### Steps for testing

1. Mark one channel as favorite
2. Archive the same channel
3. Verify it does not shown under favorite section

1. Mark one channel archived
2. Mark it as Favorite
3. Verify it should display under favorite but not under the archived/so it should be also visible under the related section (e.g Exercise)

### Screenshots

https://github.com/user-attachments/assets/4408f1b3-ea22-4037-9da2-077f37b7ad83

### DIFF
--- a/feature/metis/manage-conversations/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/manageconversations/ui/conversation/overview/ConversationOverviewViewModel.kt
+++ b/feature/metis/manage-conversations/src/main/kotlin/de/tum/informatics/www1/artemis/native_app/feature/metis/manageconversations/ui/conversation/overview/ConversationOverviewViewModel.kt
@@ -444,7 +444,14 @@ class ConversationOverviewViewModel(
 
     fun markConversationAsHidden(conversationId: Long, hidden: Boolean): Deferred<Boolean> {
         return viewModelScope.async(coroutineContext) {
-            conversationService.markConversationAsHidden(
+            // Get current conversation state to check if it's favorite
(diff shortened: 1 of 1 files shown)


FelberMartin (approved):
code lgtm

PR #670, opened 2025-07-19
`Improvement`:Arrange user setting to look more native

eylulnc:
### Problem Description
In the settings, users sometimes can find the passkey, also the design is not feeling similar to native 

### Changes
 
Instead of profile name etc, general header added again with subtitle, that indicates other action can be taken in the next step. Such as Account information -> User will see profile and option to add passkey so ( Profile, passkey)

### Steps for testing
Open app
Go to settings
See visually account and notification is clearly display and sub actions also visible as footer text

### Screenshots
Screenshot 2025-07-19 at 17 03 37


anian03 (approved):
Changes lgtm (cannot judge whether this looks more native though, not an Android user 😉)

FelberMartin (approved):

PR #671, opened 2025-07-30
`Improvement`: Display recommend update one time

eylulnc:
### Problem Description
Due to a misunderstanding the recommend update was missing for one time display when app open is recommended version is greater than current version and min version has equal or smaller than current. We display the recommend update as one time display each time app opens.

### Changes
 
Add dismissState to UpdateRepository
Arrange UpdateScreen to display recommended text

### Steps for testing
Prequisites: increase the recommended version on local server.

Open the app and verify the screen displays and dismisses on click to Skip
Stick around a little bit like ( a minute or two) to be sure the screen is not opening again.
Close the app reopen verify you see the recommend screen.

### Screenshots
Screenshot 2025-07-28 at 17 40 50

https://github.com/user-attachments/assets/818737c5-9a0e-4ac8-ae26-880cc70ae8c5


anian03 (approved):
Tested locally, both force updates and recommended updates appear to show up correctly

PR #702, opened 2026-09-15
`Chore`: E2E baseline control (do not merge)

krusche:
Control run, not for merge.

The E2E job has not executed a test since #689 on 24 July (build-logic targets JVM 21, the workflow set up JDK 17), and `continue-on-error` kept the check green throughout. #699 fixes that, and the first real run failed every test with `Login not successful: 401`.

This branch carries **only** the three infrastructure commits from #699 and none of its API changes, to establish whether those 401s are pre-existing. If E2E fails here the same way, they are.


krusche:
Served its purpose: this control run proved the E2E failures on #699 were pre-existing and not caused by its API changes — an infrastructure-only branch failed identically with 20x `Login not successful: 401`. The cause is fixed in #699 (test users were created without `internal: true`, so the server stored no password hash). Closing; nothing here to merge.

